PR #2, opened 2013-01-21
update pluginlib macro

jon-weisz:


jkammerl:
Thanks for fixing the warnings!

PR #6, opened 2013-02-20
fixes to make it comply with wiki

kalectro:
now also works without specifying publishing interval like described in the the wiki. Also renamed rate_ to time_ since it is no rate


vrabaud:
thx !

PR #8, opened 2013-03-19
fix GIT_UP with a workaround (leave a TODO) and set a working revision

MichaelKorn:
33308c1e14f0eaed9d204d00aff9c0193608dde6 don't work ROS because of  ../../lib/libpcl_io.so.1.7.0: undefined reference to `ros::TimeBase<ros::Time, ros::Duration>::fromNSec(unsigned long long)'


jkammerl:
Thanks Michael! It would be still important what causes the error. I will look into this asap.

MichaelKorn:
There is a small pull request regarding this error. With this change I can compile fuerte-unstable-devel with the newest pcl version.

jkammerl:
Thanks for the patch!

PR #11, opened 2013-05-15
revert removed directories

k-okada:


jkammerl:
Thanks for this contribution! 

I don't know why these filters have been removed a long time ago, but it's great to have them back.

PR #23, opened 2013-06-19
bag_to_pcd: check return code of transformPointCloud()

mintar:
This fixes a bug where bag_to_pcd segfaults because of an ignored 
tf::ExtrapolationException.


jkammerl:
Thanks for your bugfix, @minar!

PR #28, opened 2013-07-19
Update common.py

rlklaser:
filters min max are bounded to only -5,5 when using rosparam load and dynamic reconfigure

### DIFF
--- a/pcl_ros/cfg/common.py
+++ b/pcl_ros/cfg/common.py
@@ -8,8 +8,8 @@ from dynamic_reconfigure.parameter_generator_catkin import *;
 def add_common_parameters (gen):
   # def add (self, name, paramtype, level, description, default = None, min = None, max = None, edit_method = ""):
   gen.add ("filter_field_name", str_t, 0, "The field name used for filtering", "z")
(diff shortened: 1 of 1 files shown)


wjwwood:
I'll look at this tomorrow.

wjwwood:
@jkammerl if you have time can you review this, to see if it is reasonable? I would imagine it is fine, but I don't know what affects this might have.

jkammerl:
Hi @wjwwood, Yes, looks reasonable. In order to completely deactivate this filter by default, would it make sense to use sys.float_info.min/max as default value?

wjwwood:
I think the defaults should be within the limits? Maybe I'm misunderstanding what you mean.

wjwwood:
I don't really have any idea how this is being used internally, so I don't even know under what conditions this would "disable" the filter.

wjwwood:
@rlklaser can you rebase against master (I fixed up the travis configs, so it should pass the automated tests now).

jkammerl:
From briefly looking into the code, it looks like these are the default values for the passthrough and voxel grid filter nodelet. You are right, these default need to be within limits and therefore some large value should be fine.

PR #41, opened 2013-11-09
revert removed directories

garaemon:
The pull request by k-okada is already merged into hydro-devel, but not merged into groovy-devel
https://github.com/ros-perception/perception_pcl/pull/11

This pull request not only includes the diff of the k-okada pull request but also reverts `pcl_nodelets.xml`.

`pcl_nodelets.xml` is required to be reverted in order to use the nodelet classes from other ros launch files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,22 @@
+language:
+  - cpp
+  - python
+python:
+  - "2.7"
+compiler:
+  - gcc
+
+install:
+  - sudo sh -c 'echo "deb http://packages.ros.org/ros-shadow-fixed/ubuntu precise main" > /etc/apt/sources.list.d/ros-latest.list'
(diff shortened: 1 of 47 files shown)


wjwwood:
You will need to add a `.travis.yml` file to this pull request before the build will pass. I recommend copying the one from the hydro branch and replacing hydro with groovy:

https://github.com/ros-perception/perception_pcl/blob/hydro-devel/.travis.yml

garaemon:
added .travis.yml.
fixed CMakeLists.txt to pass the travis-ci.
fixed moving_least_squares.h because it used `pcl::Normal` which is not supported by `pcl::MovingLeastSquares`.

https://travis-ci.org/ros-perception/perception_pcl/builds/13753913

PR #45, opened 2014-04-02
Use cmake_modules to find eigen on indigo

scpeters:
Should be compatible with hydro, needed for indigo.

See also this [ros answers post](http://answers.ros.org/question/144996/find-eigen-on-os-x/) and the [ros indigo migration guide](http://wiki.ros.org/indigo/Migration#cmake_modules).


paulbovbel:
Merged into indigo-devel branch, thanks!

PR #47, opened 2014-04-07
Update to support TF2

### LINKED ISSUE #46: bag_to_pcd does not support tf2
Support for tf2 needed for Hydro and later

This is the quick way of fixing it.

Patch

```
diff --git a/pcl_ros/tools/bag_to_pcd.cpp b/pcl_ros/tools/bag_to_pcd.cpp
index b9114cf..d1ec14e 100644
--- a/pcl_ros/tools/bag_to_pcd.cpp
+++ b/pcl_ros/tools/bag_to_pcd.cpp
@@ -91,6 +91,7 @@ int

   view.addQuery (bag, rosbag::TypeQuery ("sensor_msgs/PointCloud2"));
   view.addQuery (bag, rosbag::TypeQuery ("tf/tfMessage"));
+  view.addQuery (bag, rosbag::TypeQuery ("tf2_msgs/TFMessage"));
   view_it = view.begin ();

   std::string output_dir = std::string (argv[3]);

```

If there is an example of the preferred migration approach I'd be happy to produce a patch.

LucidOne:
Fixes #46

### DIFF
--- a/pcl_ros/tools/bag_to_pcd.cpp
+++ b/pcl_ros/tools/bag_to_pcd.cpp
@@ -91,6 +91,7 @@ int
 
   view.addQuery (bag, rosbag::TypeQuery ("sensor_msgs/PointCloud2"));
   view.addQuery (bag, rosbag::TypeQuery ("tf/tfMessage"));
+  view.addQuery (bag, rosbag::TypeQuery ("tf2_msgs/TFMessage"));
   view_it = view.begin ();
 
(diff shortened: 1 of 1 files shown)


wjwwood:
@tfoote

tfoote:
Looks reasonable. We take advantage of the runtime interchangeabilty but it probably does need to be called out here when filtering the bag.

PR #66, opened 2014-10-13
Update pcl_nodelets.xml

MattDerry:
Included missing closing library tag.  This was causing the pcl/Filter nodelets below the missing nodelet tag to not be exported correctly.

### DIFF
--- a/pcl_ros/pcl_nodelets.xml
+++ b/pcl_ros/pcl_nodelets.xml
@@ -33,6 +33,7 @@
       in parallel, using the OpenMP standard.
     </description>
   </class>
+</library>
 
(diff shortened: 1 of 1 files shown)


MattDerry:
I recently installed indigo and found that the pcl/VoxelGrid and pcl/PassThrough filter nodelets were not being recognized.  Upon further inspection I noticed that the closing library tag was missing from pcl_nodelets.xml.  After applying this fix, this once again worked normally.

paulbovbel:
Thanks for the fix!

paulbovbel:
Must have gotten lost during a merge

PR #70, opened 2014-11-20
Update common.py

dani-carbonell:
Extended filter limits up to ±100000.0 in order to support intensity channel filtering.


paulbovbel:
Thanks Dani

PR #75, opened 2015-02-11
proper template specialization for pcl point cloud header access

bricerebsamen:
This is to fix ros/geometry#55
- `HasHeader` "returns false" for `pcl::PointCloud<T>`
- removed the partial specialization of `TimeStamp` and `FrameId` for `pcl::PointXYZ` and `pcl::Normal` and made it generic for all point types.

Note that the test will fail to compile until ros/geometry#80 has been merged.

### DIFF
--- a/pcl_ros/src/test/test_tf_message_filter_pcl.cpp
+++ b/pcl_ros/src/test/test_tf_message_filter_pcl.cpp
@@ -1,0 +1,390 @@
+/*
+ * Copyright (c) 2008, Willow Garage, Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *     * Neither the name of the Willow Garage, Inc. nor the names of its
+ *       contributors may be used to endorse or promote products derived from
+ *       this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ * AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/** \author Josh Faust */
+
+#include <pcl/point_types.h>
+#include <pcl/point_cloud.h>
+
+#include <tf/message_filter.h>
+#include <tf/transform_listener.h>
+#include <tf/transform_broadcaster.h>
+
+#include <boost/bind.hpp>
+#include <boost/scoped_ptr.hpp>
+
+#include <pcl_ros/point_cloud.h>
+#include <pcl_conversions/pcl_conversions.h>
+
+#include <ros/ros.h>
+
+#include <gtest/gtest.h>
+
+using namespace tf;
+
+// using a random point type, as we want to make sure that it does work with
+// other points than just XYZ
+typedef pcl::PointCloud<pcl::PointXYZRGBNormal> PCDType;
+
+
+/// Sets pcl_stamp from stamp, BUT alters stamp
+/// a little to round it to millisecond. This is because converting back
+/// and forth from pcd to ros time induces some rounding errors.
+void setStamp(ros::Time &stamp, pcl::uint64_t &pcl_stamp)
+{
+  // round to millisecond
+  static const uint32_t mult = 1e6;
+  stamp.nsec /= mult;
+  stamp.nsec *= mult;
+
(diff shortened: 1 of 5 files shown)


pcl_ros/src/test/test_tf_message_filter_pcl.cpp
```diff
@@ -0,0 +1,390 @@
...
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/** \author Josh Faust */
```
wjwwood: Probably can drop this line.

pcl_ros/src/test/test_tf_message_filter_pcl.cpp
```diff
@@ -0,0 +1,390 @@
...
+  boost::scoped_ptr<tf::MessageFilter<PCDType> > tf_filter;
+
+  spinner.start();
+  for (int i = 0; i < 3; ++i) {
+    tf_listener.reset(new tf::TransformListener());
+    // Have callback fire at high rate to increase chances of race condition
+    tf_filter.reset(
+      new tf::MessageFilter<PCDType>(*tf_listener,
+                                                         "map", 5, threaded_nh,
+                                                         ros::Duration(0.000001)));
```
wjwwood: nitpick: The indention here is pretty strange.

wjwwood:
The changes look ok to me, but some caution would be appropriate when considering how this might impact existing code. I'll leave it to @LucidOne and @paulbovbel at this point.

bricerebsamen:
I think there is a cycle between this PR and the ros/geometry#80 PR. Probably I should create separate PRs for the actual change and the corresponding test. Also I should create a separate test for tf2 message filters.

LucidOne:
I know the velodyne has a GPS clock for timestamping poointclouds so I want to test it with actual hardware before merging.

bricerebsamen:
The velodyne does have an optional gps clock input. It could be used to get a GPS referenced timestamp for each point. 

However, as per `velodyne_driver` and `velodyne_pointcloud` it's not being used. The stamp of the `VelodyneScan` message is set to when the packet is being received (as per the system clock), and the stamp of the point cloud produced by `velodyne_pointcloud` is set to the stamp of the input `VelodyneScan` message.

Btw, as a result of this, transforming a point cloud with more than a small fraction of a whole spin is going to be inaccurate if using only the header's stamp. On my side, I set the header's stamp to the stamp of the last point in the scan, and I transform the point cloud by taking into account the individual stamps for each point.

LucidOne:
I have had some meetings at work about getting per point timestamps so there was some talk about fixing this, but it looks like it has not been done yet. http://velodynelidar.com/lidar/products/manual/63-9113%20HDL-32E%20manual_Rev%20E_NOV2012.pdf (page 27)
This also reminds me that we have a velodyne GPS/IMU driver here that I need to get open sourced.

I'll compile everything tomorrow to make sure it doesn't break anything unexpected for me.

LucidOne:
Any chance you want to check if this will break ABI before I merge it into Indigo?
I think that's my last outstanding concern.

bricerebsamen:
sure, but you'll have to give some detailed instructions.

LucidOne:
I agree that this is a bug that should be fixed for Indigo, but I also don't want to merge this and have @jack-oquin or someone else tell me it broke their robot. Sorry, it is taking so long to merge.

So according to this blog it looks like template changes can break ABI compatibility
http://blogs.gentoo.org/mgorny/2012/08/20/the-impact-of-cxx-templates-on-library-abi/
ros-indigo-perception-pcl is currently at 1.2.6 so as far as I understand reading REP9 we need to keep binary compatibility.
http://www.ros.org/reps/rep-0009.html
This looks like a tool that can check ABI compatibility but I have not had a chance to test it
http://ispras.linuxbase.org/index.php/ABI_compliance_checker

bricerebsamen:
I tried using `abi-compliance-checker`

I created the following xml files:

pcl_ros_v1.2.6.xml:

```
<version>
    1.2.6
</version>

<headers>
    /opt/ros/indigo/include/pcl_ros
</headers>

<libs>
    /opt/ros/indigo/lib
</libs>
```

and pcl_ros_v1.2.7.xml

```
<version>
    1.2.7
</version>

<headers>
    /home/brice/code/workspaces/225/devel/include/pcl_ros
</headers>

<libs>
    /home/brice/code/workspaces/225/devel/lib
</libs>
```

and called `abi-compliance-checker` this way:

```
abi-compliance-checker -lib pcl_ros -old pcl_ros_v1.2.6.xml -new pcl_ros_v1.2.7.xml
```

It failed:

```
$ abi-compliance-checker -lib pcl_ros -old pcl_ros_v1.2.6.xml -new pcl_ros_v1.2.7.xml 
preparation, please wait ...
Using GCC 4.8 (x86_64-linux-gnu)
checking header(s) 1.2.6 ...
ERROR: some errors occurred when compiling headers
ERROR: see log for details:
  /home/brice/logs/pcl_ros/1.2.6/log.txt

ERROR: can't compile header(s)
```

the details in the log are:

```
In file included from /tmp/CNRzHMpir6/dump1.h:28:0:
/opt/ros/indigo/include/pcl_ros/features/normal_3d_tbb.h:44:40: fatal error: pcl/features/normal_3d_tbb.h: No such file or directory
 #include <pcl/features/normal_3d_tbb.h>
```

indeed, the file is missing:

```
$ ls /usr/include/pcl-1.7/pcl/features/normal_3d*
/usr/include/pcl-1.7/pcl/features/normal_3d.h
/usr/include/pcl-1.7/pcl/features/normal_3d_omp.h
```

Not sure why this is even possible nor what to do next

LucidOne:
Might be a CMake problem, that file should only be included if you have TBB (an OMP alternative)
`#if defined HAVE_TBB`

Maybe this will work
http://abicheck.sourceforge.net/

bricerebsamen:
the odd thing is that `#if defined HAVE_TBB` is commented in `/opt/ros/indigo/include/pcl_ros/features/normal_3d_tbb.h`

paulbovbel:
@LucidOne @bricerebsamen any thoughts on this? If it break binary compatibility, we can merge into a new branch (jade-devel), since Jade is still in beta.

LucidOne:
I'll figure this out today.

jack-oquin:
Since the change appears to affect the ABI, releasing it only in Jade seems reasonable.

LucidOne:
Thanks Jack!
I'll get this released to Jade today.

paulbovbel:
@LucidOne, ready to merge this into jade-devel?

LucidOne:
I have Monday set aside for hacking holiday to get this done.

LucidOne:
We should be ready to branch jade-devel after #91 is merged

LucidOne:
@bricerebsamen
Can you re-submit this pull request against jade-devel?

LucidOne:
Closed for jade-devel in #93

PR #80, opened 2015-03-17
Keep exact timestamp even though applying filter nodelets

garaemon:


paulbovbel:
Thank you. Could you make an accompanying pull for indigo please?

PR #85, opened 2015-03-25
Add more check before calling combineClouds and keep header information strictly

garaemon:
Current `pcl/PointCloudConcatenateDataSynchronizer` does not work.

We need more check when calling combineClouds because `if (in3)` and so on are not enough.

And also timstamp is not kept strictly after `pcl::concatenatePointCloud`, so copy timestamp after calling it.


paulbovbel:
Thank you. Could you make an accompanying pull for indigo please?

PR #91, opened 2015-05-26
Sync pcl_nodelets.xml from hydro to indigo

### LINKED ISSUE #86: Nodelet definitions
Looks like hydro-devel and indigo-devel nodelets.xml don't agree, should be reconciled.

http://answers.ros.org/question/206352/pcl_ros-nodelets-do-not-exist/

LucidOne:
Fixes to pass catkin lint -W1
http://www.cmake.org/pipermail/cmake/2011-May/044295.html
Closes #86


paulbovbel:
+1, thanks @LucidOne

PR #97, opened 2015-06-08
Remove pointcloud_to_laserscan package

paulbovbel:
No longer depends on PCL, move elsewhere for jade release.


LucidOne:
I think this is fairly widely used, I'd guess that it makes sense to have some sort of deprecation plan before just removing it.

paulbovbel:
Could add a link from the directory for Jade, for anyone trying to find the source

LucidOne:
Another option would be to have a script that prints a deprecation warning and then executes the code.
So other developers have time to update.

paulbovbel:
rospack resolution and etc. would still work fine, I'm just proposing moving the pointcloud_to_laserscan package out of the perception_pcl metapackage/repo.

The only thing that would change is where people would go to find a source, hence a simple deprecation-type README in an empty pointcloud_to_laserscan subdirectory would probably do the trick.

LucidOne:
I have not tested it but resolution makes sense.
One thing to consider is if it makes sense to have pointcloud_to_laserscan
in it's own repo. Another idea might be to talk to @idryanov and look at merging https://github.com/ccny-ros-pkg/scan_tools with pointcloud_to_laserscan into a meta-package

PR #98, opened 2015-06-08
Merge pcl_conversions and pcl_msgs into perception_pcl

paulbovbel:
For centralized issue tracking and release convenience.


paulbovbel:
@wjwwood @tfoote, any thoughts if this is reasonable, since all these packages belong to the perception_pcl metapackage anyways? I had good success doing this for the Husky robot packages recently, cuts down on the forgotten issue effect.

wjwwood:
I wouldn't do this, but if you think it will be easier for you to maintain then I guess it's ok. Just some things to consider off the top of my head:
- The git history is going to be pretty messed up.
- The version tags are going to be messed up since you don't have one version for all of the packages in the repository.
- You probably don't want to do this for anything that's already released (Jade, Indigo, Hydro, etc...).
  - If you do you'll need to level them all off to a single new version that is greater than all the previous ones and release for all affected distributions.

I would say -1, but you guys are the maintainers, so you should do what you want with it. The way I look at it, once you've got more than like 5 repositories it doesn't matter how many you have, because you need to be able to effectively context switch between GitHub repositories anyways.

paulbovbel:
@LucidOne, +/- 1?

PR #100, opened 2015-06-19
don't depend on roscore if not transforming pointclouds

jwm:
`roscore` is only necessary if `bag_to_pcd` is performing transformations. If `target_frame` isn't specified, there's no need to depend on `roscore`.

This change makes `roscore` optional if `target_frame` isn't specified. It works fine in the no-transformations case, but we don't have any bags available that require transformation. If there are test bags available somewhere, we'd be happy to test that case as well.


bmagyar (approved):

SteveMacenski:
I'm going to close this out -- if you would still like this in, more than happy to facilitate but given its 4 years old I'm guessing you don't care anymore (and from a now-`unknown repository`)

PR #102, opened 2015-06-29
pointcloud_to_pcd: Added option for fixed/custom filename

wxmerkt:
This adds the option to specify a custom filename independent of the pointcloud header timestamp (which might be useful in some scenarios where the name of the pointcloud file needs to be used as input somewhere else). It does not make use of the specified prefix. If the file exists, it will be overwritten with new pointcloud data upon receiving a new pointcloud message. [Due to this, this is only/most useful in scenarios where there is a low rate of pointclouds coming in.]

I have not found specific suggestion/guidance on how to structure/format contributions - please advice and I'll change accordingly.

Also, please close if this feature is not of interest.

Thanks for your consideration and a fantastic library and package!

### DIFF
--- a/pcl_ros/tools/pointcloud_to_pcd.cpp
+++ b/pcl_ros/tools/pointcloud_to_pcd.cpp
@@ -63,6 +63,7 @@ class PointCloudToPCD
 
   private:
     std::string prefix_;
+    std::string filename_;
     bool binary_;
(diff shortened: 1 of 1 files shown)


bmagyar (approved):

paulbovbel:
LGTM, sorry for the delay!

PR #107, opened 2015-08-18
Add fixed_frame to pointcloud_to_pcd

jspricke:


paulbovbel:
If tf(2) data is not actively being broadcast in the system, this will fail, which may be problematic for users with bags or running otherwise isolated systems. Could you follow legacy behaviour when the fixed_frame parameter is not provided?

jspricke:
Good point, fixed now.

mintar:
@paulbovbel , any news on this?

paulbovbel:
Where does the time go. LGTM, thanks for poking @mintar.

Issue #106, opened 2015-07-30
Eigen3 cmake module not found.


rcodddow:
I am try to install Jade on OS X 10.10 Yosemite. 

I am stuck trying to compile plc_ros because Eigen3 can be found

==> Processing catkin package: 'pcl_ros'
==> Building with env: '/Users/robert/ros_catkin_ws/install_isolated/env.sh'
==> cmake /Users/robert/ros_catkin_ws/src/perception_pcl/pcl_ros -DCATKIN_DEVEL_PREFIX=/Users/robert/ros_catkin_ws/devel_isolated/pcl_ros -DCMAKE_INSTALL_PREFIX=/Users/robert/ros_catkin_ws/install_isolated -DCMAKE_BUILD_TYPE=Release -G Unix Makefiles in '/Users/robert/ros_catkin_ws/build_isolated/pcl_ros'
-- Boost version: 1.58.0
-- Found the following Boost libraries:
--   system
--   filesystem
--   thread
CMake Error at CMakeLists.txt:7 (find_package):
  By not providing "FindEigen3.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Eigen3", but
  CMake did not find one.

  Could not find a package configuration file provided by "Eigen3" with any
  of the following names:

```
Eigen3Config.cmake
eigen3-config.cmake
```

  Add the installation prefix of "Eigen3" to CMAKE_PREFIX_PATH or set
  "Eigen3_DIR" to a directory containing one of the above files.  If "Eigen3"
  provides a separate development package or SDK, be sure it has been
  installed.

-- Configuring incomplete, errors occurred!
See also "/Users/robert/ros_catkin_ws/build_isolated/pcl_ros/CMakeFiles/CMakeOutput.log".
<== Failed to process package 'pcl_ros': 
  Command '['/Users/robert/ros_catkin_ws/install_isolated/env.sh', 'cmake', '/Users/robert/ros_catkin_ws/src/perception_pcl/pcl_ros', '-DCATKIN_DEVEL_PREFIX=/Users/robert/ros_catkin_ws/devel_isolated/pcl_ros', '-DCMAKE_INSTALL_PREFIX=/Users/robert/ros_catkin_ws/install_isolated', '-DCMAKE_BUILD_TYPE=Release', '-G', 'Unix Makefiles']' returned non-zero exit status 1

Reproduce this error by running:
[...]

PR #108, opened 2015-10-10
Fallback to cmake_modules if Eigen3 not found

jmtatsch:
for the time being as proposed in http://wiki.ros.org/jade/Migration#Eigen_CMake_Module_in_cmake_modules

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -4,7 +4,15 @@ project(pcl_ros)
 ## Find system dependencies
 find_package(cmake_modules REQUIRED)
 find_package(Boost REQUIRED COMPONENTS system filesystem thread)
-find_package(Eigen3 REQUIRED)
+find_package(Eigen3)
+if(NOT EIGEN3_FOUND)
+  # Fallback to cmake_modules
+  find_package(cmake_modules REQUIRED)
+  find_package(Eigen REQUIRED)
+  set(EIGEN3_INCLUDE_DIRS ${EIGEN_INCLUDE_DIRS})
+  set(EIGEN3_LIBRARIES ${EIGEN_LIBRARIES})
+  # Possibly map additional variables to the EIGEN3_ prefix.
+endif()
 find_package(PCL REQUIRED)
 
 ## Find catkin packages
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -4,14 +4,8 @@ project(pcl_ros)
 ## Find system dependencies
 find_package(cmake_modules REQUIRED)
 find_package(Boost REQUIRED COMPONENTS system filesystem thread)
-find_package(Eigen3)
-if(NOT EIGEN3_FOUND)
-  # Fallback to cmake_modules
-  find_package(cmake_modules REQUIRED)
-  find_package(Eigen REQUIRED)
-  set(EIGEN3_INCLUDE_DIRS ${EIGEN_INCLUDE_DIRS})
-  set(EIGEN3_LIBRARIES ${EIGEN_LIBRARIES})
-endif()
+find_package(PkgConfig)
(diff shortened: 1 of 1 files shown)


pcl_ros/CMakeLists.txt
```diff
@@ -4,7 +4,15 @@ project(pcl_ros)
...
 find_package(Boost REQUIRED COMPONENTS system filesystem thread)
-find_package(Eigen3 REQUIRED)
+find_package(Eigen3)
+if(NOT EIGEN3_FOUND)
+  # Fallback to cmake_modules
+  find_package(cmake_modules REQUIRED)
+  find_package(Eigen REQUIRED)
+  set(EIGEN3_INCLUDE_DIRS ${EIGEN_INCLUDE_DIRS})
+  set(EIGEN3_LIBRARIES ${EIGEN_LIBRARIES})
+  # Possibly map additional variables to the EIGEN3_ prefix.
```
wjwwood: This comment is probably not need here.

wjwwood:
Lgtm, I'll let @paulbovbel review and merge, etc.

jmtatsch:
Since I installed Eigen 3.2.6 this workaround fails with

```
CMake Error at /opt/ros/jade/share/catkin/cmake/catkin_package.cmake:159 (message):
  catkin_package() DEPENDS on 'Eigen3' which must be find_package()-ed
  before.  If it is a catkin package it can be declared as CATKIN_DEPENDS
  instead without find_package()-ing it.
  /opt/ros/jade/share/catkin/cmake/catkin_package.cmake:98 (_catkin_package)
  CMakeLists.txt:64 (catkin_package)
```

jmtatsch:
Using pkgconfig as proposed in https://github.com/ros-perception/perception_pcl/issues/106 works fine for me.

wjwwood:
@jmtatsch it fails becuase you have to either set the `Eigen3_FOUND` variable to the value of `Eigen_FOUND` (in the case that you always give `catkin_package()` the `Eigen3` name) or you have to conditionally pass the correct name (either `Eigen` or `Eigen3`) to `catkin_package()`.

paulbovbel:
@wjwwood, given this issue with OSX, would you reconsider deprecating the FindEigen.cmake in cmake_modules?

wjwwood:
@paulbovbel Which issue?

paulbovbel:
Sorry, #106. The issue being that FindEigen3.cmake is not provided by Eigen in OSX, or Arch.

wjwwood:
@paulbovbel I believe OS X has one now: https://github.com/Homebrew/homebrew/issues/35225

As for Arch, I still think it would be best for them to fix their packages, since the upstream Eigen3 release provides the module.

paulbovbel:
My mistake, I thought this was related. @jmtatsch , what's the motivation for this PR?

jmtatsch:
@paulbovbel The motivation for this PR is simply that pcl_ros does (still) not find the current Eigen on OS X El Capitan without it. Sure it would be preferred if `find_package(Eigen3 REQUIRED)` simply worked as expected on all platforms.

wjwwood:
@jmtatsch You should be trying `find_package(Eigen)` first, since that's what is distributed with Eigen. If that doesn't work then it's a bug of the Homebrew formulae and we should file an issue.

wjwwood:
Sorry, that should be `find_package(Eigen3)`. The `find_package(Eigen)` is from our `cmake_modules`.

jmtatsch:
Still an open issue as of now, here is the relevant home-brew issue https://github.com/Homebrew/homebrew/issues/44937

paulbovbel:
Apologies for the delay.

I'm extrapolating from https://github.com/ros-planning/moveit/commit/a6ba84fb98369c3710fb8a4e0adb76e211d42961 that this has been resolved. However, not having any experience with homebrew/OSX, i'm happy to take a patch if necessary.

PR #112, opened 2016-01-16
[pcl_ros] Update pcd file to publish by service call

wkentaro:


wkentaro:
ping

wkentaro:
ping

pcl_ros/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ find_package(PCL REQUIRED)
 find_package(catkin REQUIRED COMPONENTS
   dynamic_reconfigure
   genmsg
+  message_generation
```
paulbovbel: Shouldn't generate messages here. There's a pcl_msgs package for this purpose

pcl_ros/tools/pcd_to_pointcloud.cpp
```diff
@@ -89,46 +94,48 @@ class PCDGenerator
...
+        /*oneshot=*/false);
+      // service to update filename
+      service_ = private_nh_.advertiseService("update_filename", &PCDGenerator::updateFilenameCb, this);
       return (0);
     }
 
+    bool updateFilenameCb(pcl_ros::UpdateFilename::Request &req,
+                          pcl_ros::UpdateFilename::Response &res)
+    {
+      boost::mutex::scoped_lock lock(mutex_);
```
paulbovbel: No need to mutex - this node is single threaded

pcl_ros/tools/pcd_to_pointcloud.cpp
```diff
@@ -89,46 +94,48 @@ class PCDGenerator
       if (file_name_ == "" || pcl::io::loadPCDFile (file_name_, cloud_) == -1)
         return (-1);
       cloud_.header.frame_id = tf_frame_;
+      // set timer to publish cloud periodically
+      timer_ = nh_.createWallTimer(
```
paulbovbel: preference to timer would be to convert publisher to latched, and make service call trigger publishing the new point cloud, much simpler flow

paulbovbel:
@wkentaro, sorry for the delay, and thanks for the contribution. I've made some comments.

wkentaro:
@paulbovbel 
Thanks for the comments.
I have added some modifying commits. 
Could you please review them? then I will squash the commits.

paulbovbel:
Thanks for addressing the message issue.

With regards to the republish timer, this would probably be more straightforward with a latched publisher, which would update the latched message every time the service is called. Would that work for your application?

wkentaro:
> With regards to the republish timer, this would probably be more straightforward with a latched publisher, which would update the latched message every time the service is called. Would that work for your application?

The new feature of latched publishing would be useful, but I expect non-latched publishing in my application.
And I hesitate to change the current behavior of the node, so the new feature should be with conditional rosparam (ex. `~latched: True/False`), I think.

paulbovbel:
I see the old while loop now, thanks. LGTM

wkentaro:
Thanks, I squashed the commits.

paulbovbel:
Going to need a corresponding PR in pcl_msgs

wkentaro:
I have send PR ros-perception/pcl_msgs#5

wkentaro:
pign

wkentaro:
ping

wkentaro:
ping

paulbovbel:
Closing in favor of #213

PR #113, opened 2016-01-22
Remove find Eigen3.cmake as Eigen3 is a header-only library

spaghetti-:


wjwwood:
Header only doesn't mean you don't need to find_package it: https://github.com/ros-perception/perception_pcl/pull/113/files#diff-403a0045dafa306b821b9261d2a16024R31

spaghetti-:
However it does not come with a FindEigen3.cmake. I will explore a better solution.

wjwwood:
The Eigen3 configure file should be provided by Eigen:

https://github.com/ros/cmake_modules/issues/35
http://wiki.ros.org/jade/Migration#Eigen_CMake_Module_in_cmake_modules

There is an Eigen module which is provided by the cmake_modules ROS package: https://github.com/ros-perception/perception_pcl/pull/113/files#diff-403a0045dafa306b821b9261d2a16024R5

See: https://github.com/ros/cmake_modules/blob/0.3-devel/README.md

I'd recommend starting with the issue you're having rather than a proposed solution with no context.

spaghetti-:
True, did not see the cmake_modules package. My bad.

PR #114, opened 2016-02-17
Fixing typo in a ROS_INFO statement

dpiet:
Statement was incorrectly informing user that PCD would be saved as binary instead of ASCII.


paulbovbel:
LGTM, thanks!

Issue #119, opened 2016-04-21
get perception_pcl building on kinetic


jgoppert:
feature request:

I got pretty close, but ran into this error:

``` bash
Starting  >>> pcl_ros                                                            
_________________________________________________________________________________
Warnings   << pcl_ros:check /home/jgoppert/git/visim/logs/pcl_ros/build.check.006.log
** WARNING ** io features related to openni2 will be disabled
** WARNING ** io features related to pcap will be disabled
** WARNING ** io features related to png will be disabled
** WARNING ** visualization features related to openni2 will be disabled
CMake Warning (dev) at CMakeLists.txt:80 (add_dependencies):
  Policy CMP0046 is not set: Error on non-existent dependency in
  add_dependencies.  Run "cmake --help-policy CMP0046" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  The dependency target "pcl_ros_generate_messages_cpp" of target
  "pcl_ros_tf" does not exist.
This warning is for project developers.  Use -Wno-dev to suppress it.

cd /home/jgoppert/git/visim/build/pcl_ros; catkin build --get-env pcl_ros | catkin env -si  /usr/bin/make cmake_check_build_system; cd -
.................................................................................
_________________________________________________________________________________
Errors     << pcl_ros:make /home/jgoppert/git/visim/logs/pcl_ros/build.make.007.log
/usr/bin/ld: cannot find -lvtkproj4
collect2: error: ld returned 1 exit status
make[2]: *** [/home/jgoppert/git/visim/devel/.private/pcl_ros/lib/pcl_ros/convert_pointcloud_to_image] Error 1
make[1]: *** [CMakeFiles/convert_pointcloud_to_image.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
/usr/bin/ld: cannot find -lvtkproj4
collect2: error: ld returned 1 exit status
make[2]: *** [/home/jgoppert/git/visim/devel/.private/pcl_ros/lib/pcl_ros/pcd_to_pointcloud] Error 1
make[1]: *** [CMakeFiles/pcd_to_pointcloud.dir/all] Error 2
/usr/bin/ld: cannot find -lvtkproj4
collect2: error: ld returned 1 exit status
make[2]: *** [/home/jgoppert/git/visim/devel/.private/pcl_ros/lib/pcl_ros/pointcloud_to_pcd] Error 1
[...]

PR #121, opened 2016-04-21
manually remove dependency on vtkproj from PCL_LIBRARIES

jackieokay:
Bit of a horrible hack to fix a link error in pcl_ros in Xenial/Kinetic (#119). If the latest version of vtk6 gets synced into Xenial this should be rejected or removed.


jgoppert:
This works for me. +1

paulbovbel:
Merged, thanks

PR #123, opened 2016-05-04
Add missing run_depend on proj

mintar:
Without this patch, rosdep does not generate a dependency on
libproj-dev. Unless libproj-dev is installed manually, compilation for
projects that depends on pcl_ros will fail with an error like this:

```
No rule to make target '/usr/lib/x86_64-linux-gnu/libproj.so'
```

### DIFF
--- a/pcl_ros/package.xml
+++ b/pcl_ros/package.xml
@@ -58,6 +58,7 @@
   <run_depend>std_msgs</run_depend>
   <run_depend>tf</run_depend>
   <run_depend>libvtk-java</run_depend>
+  <run_depend>proj</run_depend>
 
   <test_depend>rostest</test_depend>
 


mintar:
I _hope_ this fixes the problem stated above. The problem is that the bloom-generated package for pcl_ros v1.4.0 still doesn't depend on libproj-dev. 

Steps to reproduce (only in Kinetic):
- uninstall libproj-dev
- create a ROS package that depends on pcl_ros, cd to it
- rosdep install --from-paths ./ -i -y --rosdistro kinetic
  - notice that this doesn't install libproj-dev, but it should!
- try to build your package

From the [run_depend section in REP 127](http://www.ros.org/reps/rep-0127.html#run-depend-multiple):

> The second type of <run_depend> is for transitive build dependencies. A common example is when one of your dependencies provides a header file included in some header exported by your package. Even if your package does not use that header when building itself, other packages depending on your header will require those transitive dependencies when they are built.

Since all packages depending on pcl_ros need `libproj-dev` installed at runtime (and pcl_ros needs it too), there's clearly a build_depend missing.

mintar:
@paulbovbel : Please merge and re-release.

@jacquelinekay : Probably interesting for you too, since you added the build_depends and babysit the Kinetic process. :)

jackieokay:
yeah, for context (you are probably aware of this since you've been following the issue), there is a known issue in vtk that causes issues with linking to libproj, with a fix that we are trying to get merged into 16.04 (it's already in 16.10): https://bugs.launchpad.net/ubuntu/+source/vtk6/+bug/1573234

pcl_ros doesn't explictly reference proj anywherem but vtk (which is a dependency of pcl) should bring in the dependency on proj, which is why the build_depend was added. But, I indeed neglected the run_depend.

Thanks for completing the workaround. I'm not a maintainer so I can't merge this, but +1 anyway.

paulbovbel:
Sounds like this package should be ported to package format 2 and use the build_export_depend tag.

http://wiki.ros.org/catkin/package.xml#Format_2_.28Recommended.29

mintar:
No, build_export_depend is not enough. The libproj-dev includes an .so library (which is strange for a -dev package IMO), and that needs to be there at runtime, so run_depend is correct.

Am 4. Mai 2016 22:52:16 MESZ, schrieb Paul Bovbel notifications@github.com:

> Sounds like this package should be ported to package format 2 and use
> the build_export_depend tag.
> 
> http://wiki.ros.org/catkin/package.xml#Format_2_.28Recommended.29
> 
> ---
> 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/ros-perception/perception_pcl/pull/123#issuecomment-216999107

## 

Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

mintar:
... or just "depend" is of course also okay, like you did in #124.

paulbovbel:
Sounds good, thanks for pointing that out!

mintar:
Closing in favor of #124 .

PR #124, opened 2016-05-04
Switch to package format 2

paulbovbel:
supercede #123

### DIFF
--- a/pcl_ros/package.xml
+++ b/pcl_ros/package.xml
@@ -1,4 +1,5 @@
-<package>
+<?xml version="1.0"?>
+<package format="2">
   <name>pcl_ros</name>
   <version>1.4.0</version>
   <description>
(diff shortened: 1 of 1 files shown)


mintar:
**TL;DR:** It's fine, let's merge this.

**Long version:** Ok, I just talked to @jspricke , and he pointed out the "proper" way to do it (when using package format 2):
- add a build_depend on libproj-dev (so that pcl_ros gets the header and the .so file, which is just a symlink to libproj.so.9.1.0 from libproj9, while building) 
- add a build_export_depend on libproj-dev (so that all dependent packages get headers + .so while building)
- add an exec_depend on libproj9 (so that pcl_ros + all dependent packages get the libproj.so.9.1.0 while running)

The only difference to just adding a `depend` on libroj-dev is that the solution above doesn't need libproj-dev + libproj9 at runtime, just libproj9. So it's just a bit of wasted hard drive space, both solutions work.

+1 for merging this PR.

mikeferguson:
+1 to merge and release -- this issue is blocking the release of navigation (https://github.com/ros-planning/navigation/issues/456) and I'd rather not scatter these sort of changes through the downstream packages that depend on PCL

jackieokay:
+1

@paulbovbel can you merge this and release today? Otherwise we can take care of it.

paulbovbel:
I can work on it tomorrow @jacquelinekay

atroyn:
Trying this fix locally still results in issues downstream as in https://github.com/ros-planning/navigation/issues/456

mintar:
Dang. I've just mailed with @jspricke (who's riding a train ATM and using me as a GitHub proxy). 

**The problem:** So in summary, the problem is that the vtk package is broken on Ubuntu Xenial at the moment. The maintainer was experimenting with it in Debian Unstable, and Ubuntu pulled a broken version from Debian Unstable into Xenial. It's fixed in Debian Unstable, so the _proper_ way to fix it would be to wait for Ubuntu to do a new pull.

**Proposed fix:** Since the problem is in vtk, it should be fixed in vtk, not in perception_pcl, navigation and all over the place, that just creates more problems. If we really don't want to wait until Ubuntu pulls vtk again, we could just temporarily copy the fixed vtk packages from Debian Unstable to packages.ros.org. We just need to be careful about the version numbers: Must be bigger than the current Ubuntu Xenial version, but should be smaller than the next version Ubuntu releases. So maybe just add a `.0` or so to the current Ubuntu version.

@wjwwood @dirk-thomas : Do you think you could do that?

mintar:
This is the relevant Ubuntu bug: https://bugs.launchpad.net/ubuntu/+source/vtk6/+bug/1573234

paulbovbel:
@mintar thanks for chasing this down. I've released with this PR in the meanwhile, and happy to take direction on how to correct this in perception_pcl if need be.

jackieokay:
Indeed, thanks @mintar for the explanation of the problem. Although it creates more work up front, I would prefer to import the fixed version from Yakkety instead of reproducing a workaround in all packages that depend on pcl; that way when the problem gets fixed, only one party needs to make a code change (the maintainers build farm). I will email @tfoote about the steps we need to take.

mintar:
Actually you can use the fixed packages directly from Debian instead of from Yakkety. But whatever you do, pay attention to the version numbers! If you simply take the version number from Yakkety, it's going to be a big mess!

tfoote:
I've put together a PPA here: https://launchpad.net/~tully.foote/+archive/ubuntu/backports/+packages if anyone has time to test with it that would be great. I'll try to find time over the weekend. It has amd64 and i386. I'm going to have to increment it to get the arm builds to run. I'll try to have it available on before the end of the weekend.

tfoote:
The PPA is up to date with new builds for all architectures.

tfoote:
Ok, the vtk6 version in xenial is patched: https://github.com/ros-infrastructure/reprepro-updater/issues/32 

We should not need to inject this dependency anymore. 

@jacquelinekay FYI

PR #126, opened 2016-05-17
Remove dependency on vtk/libproj-dev

mintar:
These dependencies were introduced in #124 to temporarily fix
missing / wrong dependencies in upstream vtk. This hack is no longer
necessary, since fixed vtk packages have been uploaded to
packages.ros.org (see #124 and ros-infrastructure/reprepro-updater#32).

### DIFF
--- a/pcl_ros/package.xml
+++ b/pcl_ros/package.xml
@@ -41,9 +41,6 @@
   <depend>std_msgs</depend>
   <depend>tf</depend>
   <depend>tf2_eigen</depend>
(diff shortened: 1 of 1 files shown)


mintar:
I didn't test this (didn't know how), but I'm pretty sure it should be working.

I was a bit confused about the `libvtk-java` dependency. Why was it there in the first place?

wkentaro:
This PR was tested at #145, and it passed.

PR #127, opened 2016-06-28
Add a max_num_saved parameter to pointcloud_to_pcd.

bhaskara:


pcl_ros/tools/pointcloud_to_pcd.cpp
```diff
@@ -88,6 +90,12 @@ class PointCloudToPCD
       if ((cloud->width * cloud->height) == 0)
         return;
 
+      if ((max_num_saved_ > 0) && (max_num_saved_ <= num_saved_))
+      {
+        ROS_INFO_STREAM("Saved max number of clouds (you can quit now)");
+        return;
```
SteveMacenski, line 96: Is this something you'd still like merged? Looks fine to me, but I'd like after the log statement for the program to cleanly shutdown. If its done, is should be done. 

Let me know if this isn't something you care for anymore and we  can close the PR

SteveMacenski (changes requested):

pcl_ros/tools/pointcloud_to_pcd.cpp
bhaskara: Feel free to close, thanks

PR #141, opened 2017-04-29
Add dependency on qtbase5-dev for find_package(Qt5Widgets)

wkentaro:
Testing at https://github.com/ros-perception/perception_pcl/pull/117

See https://github.com/ros-perception/perception_pcl/pull/117#issuecomment-298158272 for detail.

PR #117, opened 2016-04-02
Add travis config '.travis.yml'

wkentaro:

PR #142, opened 2017-05-02
Fix qt5 widgets

mikaelarguedas:
I got pcl_ros to compile on Debian Stretch with the following change. Not ready for merge yet, I need to test on the other platforms to be sure it's not breaking anything.

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -6,7 +6,8 @@ find_package(cmake_modules REQUIRED)
 find_package(Boost REQUIRED COMPONENTS system filesystem thread)
 find_package(Eigen3 REQUIRED)
 find_package(PCL REQUIRED)
-
+find_package(Qt5 REQUIRED COMPONENTS Widgets)
+set(QT_LIBRARIES Qt5::Widgets)
 if(NOT "${PCL_LIBRARIES}" STREQUAL "")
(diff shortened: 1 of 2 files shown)


mikaelarguedas:
tested on Xenial and Zesty, it seems to compile fine (with a few deprecation warnings). I think this is ready for review

wkentaro:
In my testing, https://github.com/ros-perception/perception_pcl/pull/141 also works, and it is also tested on travis: https://github.com/ros-perception/perception_pcl/pull/117.
Is below changes are necessary to make the build pass, or better?

```cmake
-# For debian: https://github.com/ros-perception/perception_pcl/issues/139
-find_package(Qt5Widgets QUIET)
-
+find_package(Qt5 REQUIRED COMPONENTS Widgets)
+set(QT_LIBRARIES Qt5::Widgets)
```

I still don't understand why pcl_ros depends on qt..

mikaelarguedas:
@wkentaro oh my bad I didn't see #141, so I was trying to offer an alternative. #141 works for me as well.

Regarding why qt shows up as a dependency, I think it's because vtk depends on it and (I guess) exports it as a build dependency. To be more precise, the dependency chain looks like this:
libvtk6-java => libvtk6-jni => libvtk6.3-qt => libqt5widgets5
I still think that vtk should handle that internally and not put constraint on the end user though...

Closing this in favor of #141

PR #140, opened 2017-04-28
Find Qt5Widgets to fix -lQt5::Widgets error

wkentaro:
Fix #139

PR #145, opened 2017-05-10
Remove dependency on vtk/libproj-dev

wkentaro:
> These dependencies were introduced in #124 to temporarily fix
> missing / wrong dependencies in upstream vtk. This hack is no longer
> necessary, since fixed vtk packages have been uploaded to
> packages.ros.org (see #124 and ros-infrastructure/reprepro-updater#32).

Close #126

### DIFF
--- a/pcl_ros/package.xml
+++ b/pcl_ros/package.xml
@@ -42,9 +42,6 @@
   <depend>std_msgs</depend>
   <depend>tf</depend>
   <depend>tf2_eigen</depend>
-  <depend>libvtk-java</depend>
-  <!-- libproj-dev needed due to error in vtk6 -->
-  <depend>proj</depend>
   <!-- qtbase5-dev needed due to error in qt5 -->
   <depend>qtbase5-dev</depend>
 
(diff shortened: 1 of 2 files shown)


paulbovbel:
that makes a lot of sense!

paulbovbel (approved):

pcl_ros/package.xml
```diff
@@ -42,9 +42,6 @@
   <depend>std_msgs</depend>
   <depend>tf</depend>
   <depend>tf2_eigen</depend>
-  <depend>libvtk-java</depend>
```
paulbovbel, line 45: Can we remove these hacks now as well?

https://github.com/ros-perception/perception_pcl/blob/lunar-devel/pcl_ros/CMakeLists.txt#L13

paulbovbel (changes requested):

pcl_ros/package.xml
wkentaro: I found it works.

mikaelarguedas:
that's pretty awesome, if it works it may be worth trying to also revert the Qt5 hacks from #140 and #141 ? (because I believe the Qt5 errors on stretch came from `vtk-java` in the first place)

wkentaro:
> that's pretty awesome, if it works it may be worth trying to also revert the Qt5 hacks from #140 and #141 ? (because I believe the Qt5 errors on stretch came from vtk-java in the first place)

It fails actually: https://github.com/ros-perception/perception_pcl/pull/146

mikaelarguedas:
> It fails actually: #146

too bad, thanks for trying it out!

paulbovbel (approved):

PR #146, opened 2017-05-12
Remove dependency on qt5

wkentaro:
Following #145


wkentaro:
Fixed via https://github.com/ros-perception/perception_pcl/pull/151

PR #147, opened 2017-05-30
Add samples and tests: statistical_outlier_removal, voxel_grid

wkentaro:
## Why?

To quickly understand the feature of each node, and test the validity of nodes.

## NodeLets

- statistical_outlier_removal
- voxel_grid

## Sample's Rviz

**statistical_outlier_removal**
sample_statistical_outlier_removal

**voxel_grid**
sample_voxel_grid


pcl_ros/CMakeLists.txt
```diff
@@ -176,6 +176,14 @@ target_link_libraries(convert_pcd_to_image ${Boost_LIBRARIES} ${catkin_LIBRARIES
 add_executable(convert_pointcloud_to_image tools/convert_pointcloud_to_image.cpp)
 target_link_libraries(convert_pointcloud_to_image ${Boost_LIBRARIES} ${catkin_LIBRARIES} ${Eigen3_LIBRARIES} ${PCL_LIBRARIES})
 
+## Downloads
+
+catkin_download(table_scene_lms400.pcd http://download.ros.org/data/pcl/table_scene_lms400.pcd
```
paulbovbel, line 181: if these are for testing purposes only, then move it to the testing if block, that way the file doesn't get downloaded during normal builds

paulbovbel (changes requested):

pcl_ros/CMakeLists.txt
wkentaro: This data is also used to run the sample, which works only if this package is installed from source.
So I need to write this line here.

paulbovbel (approved):
Fair enough. Thanks for the addition, apologies for the delay.

wkentaro:
Thanks for the review.

Issue #139, opened 2017-04-27
[pcl_ros] Qt5::Widgets not found on Debian stretch


mikaelarguedas:
pcl_ros binary jobs fail at linktime: `/usr/bin/ld: cannot find -lQt5::Widgets` 
Job link: http://build.ros.org/view/Lbin_ds_dS64/job/Lbin_ds_dS64__pcl_ros__debian_stretch_amd64__binary/2/console
timestamp:"13:17:33"

wkentaro:
I've tried to fix this, but still not sure how to install Qt5::Widget library.

paulbovbel:
maybe https://packages.debian.org/search?keywords=libqt5widgets5?

the real question is why that fails only for debian stretch....

wkentaro:
libqt5widgets5 is installed, but it still fails

```
......................................................................................................
______________________________________________________________________________________________________
Errors     << pcl_ros:make /root/catkin_ws/logs/pcl_ros/build.make.000.log
/usr/bin/ld: cannot find -lQt5::Widgets
collect2: error: ld returned 1 exit status
make[2]: *** [/root/catkin_ws/devel/.private/pcl_ros/lib/pcl_ros/convert_pcd_to_image] Error 1
make[1]: *** [CMakeFiles/convert_pcd_to_image.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
/usr/bin/ld: cannot find -lQt5::Widgets
collect2: error: ld returned 1 exit status
make[2]: *** [/root/catkin_ws/devel/.private/pcl_ros/lib/pcl_ros/convert_pointcloud_to_image] Error 1
make[1]: *** [CMakeFiles/convert_pointcloud_to_image.dir/all] Error 2
make: *** [all] Error 2
cd /root/catkin_ws/build/pcl_ros; catkin build --get-env pcl_ros | catkin env -si  /usr/bin/make --jobserver-fds=6,7 -j; cd -
......................................................................................................
Failed     << pcl_ros:make                         [ Exited with code 2 ]
Failed    <<< pcl_ros                              [ 1 minute and 23.6 seconds ]
[build] Summary: 1 of 2 packages succeeded.
[build]   Ignored:   1 packages were skipped or are blacklisted.
[build]   Warnings:  1 packages succeeded with warnings.
[build]   Abandoned: None.
[build]   Failed:    1 packages failed.
[build] Runtime: 1 minute and 33.6 seconds total.
[...]

PR #151, opened 2017-07-15
Qt5 dependency

mikaelarguedas:
@wkentaro @paulbovbel I'm opening this for visibility and share my findings when looking into the Qt dependency issue and am open to discussion to reduce the scope or close it if it's not considered of interest.
This is an alternative to https://github.com/ros-perception/perception_pcl/pull/146

I looked a bit more into this issue #139:
to answer the [comment](https://github.com/ros-perception/perception_pcl/pull/142#issuecomment-298602081):
> I still don't understand why pcl_ros depends on qt..

It is because PCL depends on vtk6 rendering tools including the Qt ones. Given that `pcl_ros` finds **all** PCL_LIBRARIES, as a result, `pcl_ros` libraries link against all vtk libraries used by PCL, including the ones depending on Qt.
To avoid depending and linking against libraries you don't use, it is good practice to `find_package()` only the components that you use. See the [first commit](https://github.com/mikaelarguedas/perception_pcl/commit/f968425079a292c3fc64829150c583bea9af2411) of this PR for a suggestion.

This change would **not** allow to get rid of the Qt5Widgets dependency though because the PCL `io` module exports dependencies on Qt5::Widgets (not sure why :confused:)
Looking a bit further into the libraries reported by the PCL IO module, it looks like several of them use Qt: see https://github.com/mikaelarguedas/perception_pcl/commit/54d94db9ae739f8bc8b81abde1f89b22490ad69a for a list.

As Far A I Understand, Ubuntu knows it is a problem and decided to make `libpcl-dev` [depend on](https://packages.ubuntu.com/zesty/libpcl-dev) `libvtk6-qt-dev` rather than fixing the root of the problem. That is why this problem doesn't appear on Ubuntu but only on debian Stretch ([no dependency on](https://packages.debian.org/stretch/libpcl-dev) `libvtk6-qt-dev`).

Possible solutions: 
 - keep the Qt5Widgets import workaround adding an unused dependency
 - merge this PR that allows to get rid of the dependency on Qt but make the CMake code a bit more complex.

In the meantime, I'll try to find out why even the io module depends on Qt and where to report the issue to get it fixed upstream

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -12,7 +12,6 @@ if(NOT "${PCL_LIBRARIES}" STREQUAL "")
   list(REMOVE_ITEM PCL_LIBRARIES "/usr/lib/libmpi.so")
 
   # For debian: https://github.com/ros-perception/perception_pcl/issues/139
-  list(REMOVE_DUPLICATES PCL_LIBRARIES)
   list(REMOVE_ITEM PCL_LIBRARIES
     "vtkGUISupportQt"
     "vtkGUISupportQtOpenGL"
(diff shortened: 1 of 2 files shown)


wkentaro:
I think removing the unused dependency is better solution, even though it requires the complex cmake code.

wkentaro (approved):

PR #153, opened 2017-08-08
Inherit NodeletLazy for pipeline with less cpu load

wkentaro:
## Why?

Checking number of subscribers in the callback is not enough to reduce cpu load with nodelet pipeline. 
If a leaf nodelet subscribes the parent node's topic, the parent node processes in the callback because `pub_output.getNumSubscribers() == 1`.

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -76,6 +76,7 @@ catkin_package(
     dynamic_reconfigure
     message_filters
     nodelet
+    nodelet_topic_tools
     pcl_conversions
     pcl_msgs
     rosbag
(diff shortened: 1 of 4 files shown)


wkentaro:
Please try using `sample_voxel_grid.launch` and enable/disable the rviz display of voxel grid output, with watching `rosnode info /voxel_grid`.
The `/voxel_grid` node only subscribes the input point cloud topic when it is being subscribed (by `rviz`).

paulbovbel (approved):
Neat! LGTM assuming tests pass

furushchev:
I'm by change doing the same work and there seems to be a bit difference each other. ( #154 )
Please compare and let us make it better. :+1: 
@wkentaro @paulbovbel

wkentaro:
The nodelet not found error is caused by mistaken syntax error of package.xml.
It is reported at https://github.com/ros/rospack/issues/77 and now it is resolved.

PR #161, opened 2017-08-22
Enhance .travis.yml

wkentaro:


paulbovbel:
Does the docker script end up rebuilding the same artifacts several times? In that case I imagine ccache would be useful.

Please try to provided at least a minimal description with your pull requests.

wkentaro:
- sudo: false: for faster booting on Travis
- ccache: in the testing script, it build twice with/without --install option, and also ccache can be cached with `cache: ccache` option in .travis.yml

.travis.yml
```diff
@@ -1,6 +1,11 @@
-sudo: required
+sudo: false
 dist: trusty
 language: generic
+services:
+  - docker
+cache:
```
paulbovbel: It looks like setting:

`cache: ccache`

Will automatically install/add to path ccache, and cache the correct directory.

https://docs.travis-ci.com/user/caching/

paulbovbel (changes requested):

.travis.yml
paulbovbel: EDIT: ah but perhaps it won't pass that to the docker container. So can just change:

```
cache:
  directories:
    - $HOME/.ccache
```

to `cache: ccache`

wkentaro: Ok, I updated.

paulbovbel (approved):

PR #163, opened 2017-08-28
remove hack now that upstream pcl has been rebuilt

mikaelarguedas:
new version of pcl has [been released](https://bugs.launchpad.net/ubuntu/+source/pcl/+bug/1704459) so this hack is not necessary anymore


wkentaro (approved):
LGTM!

PR #184, opened 2018-04-08
back to find_package(PCL ..) section to original code in 1.4.1

k-okada:
Since 1.4.2, due to the code is introduced in https://github.com/ros-perception/perception_pcl/pull/151, it removes `libpcl_visualization` and/or `libpcl_recognitoin` from `pcl_ros_LIBRARIES` and cause regression (ex. https://github.com/jsk-ros-pkg/jsk_recognition/issues/2259) 
Hope we should have same behavior as 1.4.1 in at least kinetic.

As for a lunar, I do not sure what should we do, but at least, I hope you'll provide all PCL components to the downstream packages
```
find_package(PCL REQUIRED COMPONENTS common
kdtree
octree
search
io
sample_consensus
filters
features
segmentation
surface
registration
recognition
keypoints
visualization
people
outofcore
tracking
)
```

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -5,7 +5,7 @@ project(pcl_ros)
 find_package(cmake_modules REQUIRED)
 find_package(Boost REQUIRED COMPONENTS system filesystem thread)
 find_package(Eigen3 REQUIRED)
-find_package(PCL REQUIRED COMPONENTS core features filters io segmentation surface)
+find_package(PCL 1.7 REQUIRED)
 
 if(NOT "${PCL_LIBRARIES}" STREQUAL "")
   # For debian: https://github.com/ros-perception/perception_pcl/issues/139


pcl_ros/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ project(pcl_ros)
 find_package(cmake_modules REQUIRED)
 find_package(Boost REQUIRED COMPONENTS system filesystem thread)
 find_package(Eigen3 REQUIRED)
-find_package(PCL REQUIRED COMPONENTS core features filters io segmentation surface)
+find_package(PCL 1.7 REQUIRED)
```
wkentaro, line 8: Why do you specify 1.7?

k-okada: 1. kinetic supports pcl 1.7 http://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021
2. origonall pcl_ros 1.4.1 uses pcl 1.7 https://github.com/ros-perception/perception_pcl/blob/1.4.1/pcl_ros/CMakeLists.txt#L8

wkentaro: Ok, so `find_package(PCL REQUIRED)` is fine, right?
I don't think specifying version is not a good idea, because it fobids use of PCL 1.8.
and FYI, in pcl 1.4.0, there was no specification of the version number.

k-okada: it should be ok, but I'm not sure. Can you create new  PR with  `find_package(PCL REQUIRED)` ?

wkentaro: Yes

wkentaro:
Closed via https://github.com/ros-perception/perception_pcl/pull/185

PR #192, opened 2018-04-30
Use industrial_ci for travis testing

paulbovbel:
This might be a nicer option, I've been using industrial_ci for most of my other packages since it keeps all the travis/testing logic external to the repo under test.

PR #193, opened 2018-05-01
Fix melodic build

clalancette:
These 2 commits seem to fix the melodic dev build, that failed here: http://build.ros.org/view/Mdev/job/Mdev__perception_pcl__ubuntu_bionic_amd64/1/consoleFull

### DIFF
--- a/pcl_conversions/CMakeLists.txt
+++ b/pcl_conversions/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(pcl_conversions)
 
-find_package(catkin REQUIRED COMPONENTS)
+find_package(catkin REQUIRED COMPONENTS pcl_msgs roscpp sensor_msgs std_msgs)
 
 find_package(PCL REQUIRED COMPONENTS common io)
 find_package(Eigen3 REQUIRED)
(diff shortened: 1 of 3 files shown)


paulbovbel:
Thanks @clalancette. What do you think of this approach (https://github.com/ros-perception/perception_pcl/pull/192/commits/7b6e086113e780a9e9617935e99b4003f92d8b24)?

I was trying to keep the dependencies transitive, since pcl_conversions is header only outside of tests.

mikaelarguedas:
It would be great if we could make it work without escalating all the dependencies to build_depend and exec_depend and keeping them transitive instead.
have you tried the following approaches?
- make the find_package of the test depends REQUIRED and pass their LIBRARIES and INCLUDE_DIRS to the tests (e.g find_package(std_msgs REQUIRED) ... include_directories(... ${std_msgs_INCLUDE_DIRS}" ...) etc
- find the test dependencies as REQUIRED catkin components to keep using the `catkin_LIBRARIES` 
 `catkin_INCLUDE_DIRS` macros for the tests

paulbovbel:
Opened a PR for the commit above here (https://github.com/ros-perception/perception_pcl/pull/194)

wkentaro:
Closed via https://github.com/ros-perception/perception_pcl/pull/194

PR #196, opened 2018-05-02
Fix a bug building on artful.

clalancette:
This should fix the errors encountered [here](http://build.ros.org/view/Mbin_uA64/job/Mbin_uA64__pcl_ros__ubuntu_artful_amd64__binary/5/consoleFull).  The comment explains it in more detail.  While we are at it, fix the link in the README.rst to point to the right repository.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

### DIFF
--- a/pcl_conversions/CMakeLists.txt
+++ b/pcl_conversions/CMakeLists.txt
@@ -1,10 +1,20 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.7)
 project(pcl_conversions)
 
 find_package(catkin REQUIRED COMPONENTS)
 
 find_package(PCL REQUIRED COMPONENTS common io)
 find_package(Eigen3 REQUIRED)
+
+# There is a bug in the Ubuntu Artful (17.10) version of the VTK package,
+# where it includes /usr/include/*-linux-gnu/freetype2 in the include
+# directories (which doesn't exist).  This filters down to the PCL_INCLUDE_DIRS,
+# and causes downstream projects trying to use these libraries to fail to
+# configure properly.  Here we remove those bogus entries so that downstream
+# consumers of this package succeed.
+if(NOT "${PCL_INCLUDE_DIRS}" STREQUAL "")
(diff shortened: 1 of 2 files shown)


paulbovbel (approved):
Thanks @clalancette!

clalancette:
Would you mind doing a new release of this for Melodic?  It will unblock a few downstream users.  Thanks!

pcl_conversions/CMakeLists.txt
```diff
@@ -1,11 +1,21 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.7)
```
mikaelarguedas, line 1: @clalancette Why 3.7 here ?
The FILTER feature is available since [3.6.3](https://cmake.org/cmake/help/v3.6/command/list.html)

clalancette: I picked 3.7 because its the oldest version of CMake that Melodic is targeting (Debian Stretch).  I wouldn't have any objections to lowering it to 3.6.3, but there's probably not much benefit since this is a melodic-devel branch anyway.

mikaelarguedas: It restricts anyone building from source on a system that has 3.6.3 for no real justification.

If we want to match the lowest in Melodic it would be 3.7.2 that's why I was wondering where 3.7 came from as it was in between, so allowing users having 3.7.0 to compile this package but not people have 3.6.3.

mikaelarguedas:
See follow-up PR https://github.com/ros-perception/perception_pcl/pull/198 to fix the CMake version and updating the links in the package.xml to point to the right repo

wkentaro:
I will backport this to kinetic in order to resolve the issue https://github.com/ros-perception/perception_pcl/issues/172

wkentaro:
Well, I found that #172 is about pcl_ros but this PR is about pcl_conversions.
So they are two different issues, so I will see the cmake of pcl_ros.

PR #199, opened 2018-05-07
Increase limits on CropBox filter parameters

james-ward:
Min and max of CropBox filter was +/- 5m. For a pointcloud from a Velodyne, for example, this is not enough.
Increased to +/- 1000m.


wkentaro (approved):

wkentaro:
Oops, it must be targeted to melodic-devel. so I will revert this in https://github.com/ros-perception/perception_pcl/pull/200.

PR #207, opened 2018-05-16
Filter include dirs that does not exist

wkentaro:
Close https://github.com/ros-perception/perception_pcl/issues/172
Close https://github.com/ros-perception/perception_pcl/issues/206


paulbovbel (approved):
VTK continues causing trouble. I wonder if there's a way to pull in PCL without pulling in any of the gui components.

It may make sense to do a double `foreach` loop here, but I'm not super picky about DRY since it's only two items.

wkentaro:
Thanks. I updated to use nested `foreach`.

> VTK continues causing trouble. I wonder if there's a way to pull in PCL without pulling in any of the gui components.

Yeah. It's caused by the `io` components of PCL, and currently I can't find the way to include the `io` component without including vtk.

mikaelarguedas:
> VTK continues causing trouble. I wonder if there's a way to pull in PCL without pulling in any of the gui components.

Unfortunately I dont think there is a way ATM as some of the PCL components used in this package are pulling in the vtk libraries (this branch is actually pulling all of PCL).
There was a similar discussion in https://discourse.ros.org/t/pcl-ros-with-pcl-1-8-ros-lunar-and-newer-means-linking-to-139-vtk-dynamic-libraries/4433/2
It may be worth opening an issue on the PCL project to see if they'd be willing to split the modules currently using VTK in two: 1 with only the PCL features without VTK and 1 with the VTK parts.

wkentaro:
I opened an issue for the vtk problem: https://github.com/ros-perception/perception_pcl/issues/208

PR #209, opened 2018-05-17
pcl_ros: fix exported includes in Ubuntu Artful

hrnr:
pcl_ros needs the same fix as pcl_conversions

This blocks building packages depending on pcl_ros for melodic on Ubuntu Artful. The same fix is already in pcl_conversions. pcl_ros needs new release for melodic.

#207 fixed this issue for kinetic, maybe we should do the same for lunar too.


clalancette:
This looks good to me; assuming the maintainers agree, we'll need to get this merged, tagged, and released to unblock downstream consumers in Melodic.  Thanks for the fix!

wkentaro (approved):

Issue #212, opened 2018-08-22
Can we make pcd_to_pointcloud publish once using a latched topic instead?


moooeeeep:
I want to publish a large, never changing pointcloud. Currently, with `pcd_to_pointcloud`, the pointcloud is only published once, which means that I have to relaunch the node whenever the node that processes it restarts. I find this inconvenient. 

Luckily the fix is easy. Just make the publisher advertise to latch the topic. Maybe I can make a PR later.

Any opinions on this?

paulbovbel:
That sounds reasonable, PRs welcome

PR #213, opened 2018-08-23
rewrote pcd_to_pointcloud to publish using a latched topic

moooeeeep:
Addressing issue https://github.com/ros-perception/perception_pcl/issues/212

I ended up to rewrite the whole thing.

 * pointcloud is published as a latched topic now
 * added functionality: ability to set config via parameters
 * shouldn't break old usage patterns
 * changed some of the output text
 * I think the code is more maintainable and overall much nicer now

Probably this will not play well with other pending PRs for this node, e.g., https://github.com/ros-perception/perception_pcl/pull/112 might need to be adapted


paulbovbel:
Generally LGTM. Looks like the test failures are unrelated...

Since @wkentaro has #112 open, maybe he could take a look at this as well.

paulbovbel:
Thanks, sorry for the delay.

PR #214, opened 2018-09-30
Wait for transform in transformPointCloud().

skohlbr:
Just wanted to add this myself, but found this fork by @shuntaraw adding a wait for transform in transformPointCloud(). The fact that there is no wait for transform so far means that it depends on sheer luck (and some factors like system load etc.) wether data can be transformed or not when another input frame is selected for the incoming cloud. It's kind of fascinating that no one ran into this in the last 6 years :)


peci1:
Thumbs up for this one. The failed tests seem to be more CI-infrastructure failures. 

@ros-pull-request-builder retest this please

paulbovbel:
LGTM!

paulbovbel (approved):

PR #215, opened 2018-10-01
Fix #150: Allow bag_to_pcd to properly respond to topic argument

### LINKED ISSUE #150: bag_to_pcd ignores the <topic_name> option
Specifying the <topic_name> option has no effect. It seems to just grab all the messages of type PointCloud2. There are only 4 messages in my log. If I invent a channel name `/foo` which doesn't exist in my log then it still grabs the four messages. Is there a bug in ignoring the topic name that was specified?

> Saving recorded sensor_msgs::PointCloud2 messages on topic /foo to test_cards
Got 780383 data points in frame tote_unpack with the following fields: x y z
Data saved to test_cards/1499971977.886862538.pcd
Got 1034334 data points in frame tote_unpack with the following fields: x y z
Data saved to test_cards/1499971977.895911841.pcd
Got 830770 data points in frame tote_unpack with the following fields: x y z
Data saved to test_cards/1499971977.910123129.pcd
Got 769933 data points in frame tote_unpack with the following fields: x y z
Data saved to test_cards/1499971977.917675317.pcd

jarvisschultz:
Argument was ignored before, as reported in #150 

Also added some more helpful printout information in case user provides wrong topic name or type.

### DIFF
--- a/pcl_ros/tools/bag_to_pcd.cpp
+++ b/pcl_ros/tools/bag_to_pcd.cpp
@@ -98,17 +98,24 @@ int
       topic_list[ci->topic] = ci->datatype;
       if (ci->topic == argv[2])
       {
-          target_topic = std::string (argv[2]);
-          view.addQuery (bag, rosbag::TopicQuery (target_topic));
+		  if (ci->datatype == std::string("sensor_msgs/PointCloud2"))
+		  {
+			  target_topic = std::string (argv[2]);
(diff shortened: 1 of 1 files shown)


paulbovbel:
LGTM, thanks! CI failures due to unsupported platforms that have been removed on master.

paulbovbel (approved):

PR #221, opened 2019-02-01
use <chrono> and <thread> on Windows

kejxu:
`usleep` is only available on Linux, use `this_thread::sleep_for` from STL on Windows to work around the issue

### DIFF
--- a/pcl_ros/tools/pcd_to_pointcloud.cpp
+++ b/pcl_ros/tools/pcd_to_pointcloud.cpp
@@ -49,6 +49,8 @@
 #include <pcl/io/pcd_io.h>
 #include <pcl/point_types.h>
 #include <pcl_conversions/pcl_conversions.h>
+#include <chrono>
+#include <thread>
 
 #include "pcl_ros/publisher.h"
 
@@ -116,7 +118,7 @@ class PCDGenerator
           continue;
         }
 
-        usleep (interval);
+        std::this_thread::sleep_for(std::chrono::microseconds(static_cast<uint32_t>(interval)));
 
         if (interval == 0)	// We only publish once if a 0 seconds interval is given
 				{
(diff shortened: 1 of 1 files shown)


kejxu:
this change is only limited to Windows build environment, should not affect Linux build result

seanyen:
@paulbovbel @wkentaro  Hope I am pinging the correct maintainers. We'd love some feedbacks and see how we can make this fix back to the upstream. Thanks! (BTW, this is verified in http://aka.ms/ros project.)

LucidOne (changes requested):
Melodic targets C++14 so I don't think this actually needs to be `#ifdef`ed
http://www.ros.org/reps/rep-0003.html#melodic-morenia-may-2018-may-2023
http://wiki.ros.org/melodic/Migration

It isn't a blocker to me for this, but if windows specific PRs are landing, I'd like to see a PR fior Travis to handle windows builds.

kejxu:
> Melodic targets C++14 so I don't think this actually needs to be `#ifdef`ed

My apologies for the delay, and thanks a ton for pointing this out! I have removed the Windows-specific guard.
 
> It isn't a blocker to me for this, but if windows specific PRs are landing, I'd like to see a PR for Travis to handle windows builds.

at this point we are having a [daily build](https://ros-win.visualstudio.com/ros-win/_build?definitionId=19
) to verify Windows build quality. We are still working on an Azure-Pipelines-based pull request builder solution, and we will propose corresponding change when that is finished =)

kejxu:
travis builds successfully for melodic, fails for kinetic builds

kejxu:
just a friendly ping @LucidOne =), any update on this?

LucidOne:
I'm good with merging this unless @paulbovbel  has any  feedback. I was just hoping to check on the failed tests first. If it isn't merged by Monday ping me again.

kejxu:
@LucidOne my apologies for the spam, but let me just send out this friendly ping again =)

LucidOne (approved):

LucidOne:
Thanks for your patience. I should have builds working again soon. I think it's an issue with rosdep network connection timing out.

kejxu:
> Thanks for your patience. I should have builds working again soon. I think it's an issue with rosdep network connection timing out.

No worries, thanks a lot for revisiting this! Hopefully the build would get fixed soon =)

PR #223, opened 2019-02-28
Catch tf::TransformException in transformPointCloud

efernandez:
This allows to catch all exceptions, not only `tf::LookupException` and `tf::ExtrapolationException`.

Note that most of the code using `transformPointCloud` already (implicitly) assumes this function doesn't throw any exception, and they check the return value, which is `false` when an exception happened (or something went wrong).

Only this code wasn't checking the return value, so I've also fixed that:
https://github.com/ros-perception/perception_pcl/blob/12482e555dfc295c47dba1e7c14560870c6a6d0e/pcl_ros/src/pcl_ros/io/concatenate_data.cpp#L214-L222

### DIFF
--- a/pcl_ros/include/pcl_ros/impl/transforms.hpp
+++ b/pcl_ros/include/pcl_ros/impl/transforms.hpp
@@ -105,12 +105,7 @@ transformPointCloudWithNormals (const std::string &target_frame,
   {
     tf_listener.lookupTransform (target_frame, cloud_in.header.frame_id, fromPCL(cloud_in.header).stamp, transform);
   }
-  catch (tf::LookupException &e)
-  {
-    ROS_ERROR ("%s", e.what ());
-    return (false);
-  }
-  catch (tf::ExtrapolationException &e)
+  catch (const tf::TransformException &e)
   {
     ROS_ERROR ("%s", e.what ());
     return (false);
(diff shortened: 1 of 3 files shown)


efernandez:
``` bash
Errors     << pcl_conversions:cmake /home/travis/catkin_ws/logs/pcl_conversions/build.cmake.000.log
** WARNING ** io features related to ensenso will be disabled
** WARNING ** io features related to davidSDK will be disabled
** WARNING ** io features related to dssdk will be disabled
** WARNING ** io features related to pcap will be disabled
** WARNING ** io features related to png will be disabled
** WARNING ** io features related to libusb-1.0 will be disabled
CMake Error at /opt/ros/melodic/share/catkin/cmake/catkinConfig.cmake:83 (find_package):
  Could not find a package configuration file provided by "roscpp" with any
  of the following names:
    roscppConfig.cmake
    roscpp-config.cmake
  Add the installation prefix of "roscpp" to CMAKE_PREFIX_PATH or set
  "roscpp_DIR" to a directory containing one of the above files.  If "roscpp"
  provides a separate development package or SDK, be sure it has been
  installed.
  CMakeLists.txt:35 (find_package)
```

doesn't look like a legit error from this PR :thinking:

paulbovbel (approved):
LGTM, thanks!

PR #195, opened 2018-05-02
Use industrial_ci for travis testing

wkentaro:
Rebased version of #192

PR #227, opened 2019-04-05
Fix Travis CI builds

LucidOne:
mkdir ~/catkin_ws/logs manually to resolve travis build errors

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -79,6 +79,7 @@ travis_time_start setup.script
 #script:
 source /opt/ros/$ROS_DISTRO/setup.bash
 find /opt -name roscppConfig.cmake
+apt-cache search '^ros-\w*-roscpp$' | cut -d\  -f1 | xargs -n1 apt-cache show
 cd ~/catkin_ws
 catkin build -p1 -j1
 catkin run_tests -p1 -j1
--- a/.travis.sh
+++ b/.travis.sh
@@ -79,7 +79,7 @@ travis_time_start setup.script
 #script:
 source /opt/ros/$ROS_DISTRO/setup.bash
 find /opt -name roscppConfig.cmake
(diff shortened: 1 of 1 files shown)


LucidOne:
This pull request is in progress

LucidOne:
```
CMake Error at /opt/ros/melodic/share/catkin/cmake/catkinConfig.cmake:83 (find_package):
  Could not find a package configuration file provided by "roscpp" with any
  of the following names:
    roscppConfig.cmake
    roscpp-config.cmake
```

It looks non-deterministic so I think the issue is rosdep is timing out on network connectivity

paulbovbel:
I would suggest poking at industrial_ci (https://github.com/ros-perception/perception_pcl/pull/195) instead, as that's been pretty reliable in the past.

LucidOne:
I'm 100% in favor of not running docker if we don't have to.
@wkentaro , do you want to rebase #195 or should I fork from `travis-industrial-rebased`?

paulbovbel:
Melodic targets bionic which Travis does not support natively, so you're kind of stuck with docker.

industrial_ci is used quite extensively throughout the ROS ecosystem and 'works'.

The alternative is to drop Travis and rely on buildfarm PR builds exclusively.

LucidOne:
Thanks. I'll see what I can come up with. Any thoughts on https://builds.sr.ht/
https://man.sr.ht/builds.sr.ht/compatibility.md

paulbovbel:
> Any thoughts on https://builds.sr.ht/ https://man.sr.ht/builds.sr.ht/compatibility.md

Not really, other than we already have two pre-baked solutions for CI tests for a ROS package, and I'd be keen to avoid NIH given limited bandwidth for upkeep:
- ros_buildfarm
- industrial_ci via travis

PR #233, opened 2019-04-26
Switch to industrial_ci

paulbovbel:
Will also re-enable ROS PR builder, as the include_directories change will suppress warnings coming from upstream pcl headers (https://github.com/ros/rosdistro/pull/21035).


LucidOne (approved):
Not sure it is important at the moment, but as an alternative to making melodic global...
```
env:
  matrix:
    - ROS_DISTRO=melodic DOCKER_BASE_IMAGE=debian:stretch
    - ROS_DISTRO=melodic DOCKER_BASE_IMAGE=ubuntu:bionic
```

Otherwise, LGTM :+1:

PR #238, opened 2019-06-26
[Windows][melodic-devel] Fix Boost linkage issue and binary install location.

seanyen:
* On Windows, `Boost.Iostreams` usage implies to link `boost_iostreams-*.lib` at linkage time, so it needs to be added explicitly in `CMakeLists.txt`.
* Follow `catkin` guidance to install `.DLL` files to the correct location: http://docs.ros.org/melodic/api/catkin/html/howto/format2/building_libraries.html


paulbovbel (approved):
Passes CI so sounds good to me. No experience building this under windows.

PR #239, opened 2019-06-26
Split off pcl_ros_filter into separate library

### LINKED ISSUE #16: Filter class should be compiled separately from pcl_ros_filters library
Documentation about the pcl_ros::Filter class http://ros.org/doc/groovy/api/pcl_ros/html/classpcl__ros_1_1Filter.html makes it appear that it is a class which can be used by implementers of new pcl filter nodelets.

However as pointed out in #9 the CMakeLists.txt file includes the line "class_loader_hide_library_symbols(pcl_ros_filters)", and pcl_ros_filters is the only library containing the compiled contents of pcl_ros::Filter.  Therefore pcl_ros::Filter is not actually usable by outside programmers.  If it is really intended to be only an internal helper class for the pcl_ros_filters library, then it should be documented as such.

However it appears to be a more generally useful helper class, making it easy to write PCL filter nodelets without a bunch of boilerplate.  To make this possible, a new library should be defined, say pcl_ros_filter or pcl_ros_filter_base which contains only pcl_ros::Filter.  That library would naturally not have class_loader_hide_library_symbols() because it would not be defining a plugin.  Then implementors of new filters would link against that library, as would the existing pcl_ros_filters library.

In the meantime the solution to my problem is to copy pcl_ros/filters/filter.cpp into my own code and do a bunch of renaming.

paulbovbel:
Resolve #16

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -87,8 +87,7 @@ generate_dynamic_reconfigure_options(
 catkin_package(
   INCLUDE_DIRS include
   LIBRARIES
-    pcl_ros_filters
-    pcl_ros_io
+    pcl_ros_filter
     pcl_ros_tf
   CATKIN_DEPENDS
     dynamic_reconfigure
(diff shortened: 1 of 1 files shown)


pcl_ros/CMakeLists.txt
```diff
@@ -87,8 +87,7 @@ generate_dynamic_reconfigure_options(
 catkin_package(
   INCLUDE_DIRS include
   LIBRARIES
-    pcl_ros_filters
```
paulbovbel, line 90: Both pcl_ros_filters and pcl_ros_io have their symbols stripped, so it makes no sense to export these libraries.

PR #240, opened 2019-07-13
Make minor fixes and cleanups

mvieth:
- Add c++11 compile option because a for loop in that package uses the c++11 standard
- Fix the bug that -1 might be used as an index
- Remove unnecessary includes

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8)
 project(pcl_ros)
 
+add_compile_options(-std=c++11)
+
 ## Find system dependencies
 find_package(cmake_modules REQUIRED)
 find_package(Boost REQUIRED COMPONENTS system filesystem thread)
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 2.8)
 project(pcl_ros)
 
-add_compile_options(-std=c++11)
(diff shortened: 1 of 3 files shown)


mvieth:
That failing check is not because of the changes I made, is it?

pcl_ros/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 2.8)
 project(pcl_ros)
 
+add_compile_options(-std=c++11)
```
paulbovbel: Not sure this is necessary, since melodic targets bionic, where gcc defaults to 14 AFAIK

Will have to look and see how other ROS packages handle this

paulbovbel:
Looks like a flaky test

SteveMacenski:
@mvieth can you remove the C++11 flag so we can ship this?

mvieth:
@SteveMacenski I added it because a C++11 style `for` loop is used [here](https://github.com/ros-perception/perception_pcl/blob/melodic-devel/pcl_ros/tools/bag_to_pcd.cpp#L96). That should be reflected in the CMakeLists, no?

SteveMacenski:
Melodic + 18.04 defaults to C++14 so you'd actually be downgrading

mvieth:
Ok, I see. So would you like me to replace it with a C++14 flag to make that explicit or remove it altogether?

SteveMacenski:
C++14 flag would be fine :+1:

SteveMacenski (approved):

SteveMacenski:
@paulbovbel any problem here now?

paulbovbel:
LGTM

PR #247, opened 2019-10-30
General cleanup

mvieth:
- Remove unnecessary includes (as recommended by include-what-you-use)
- Comment out unused parameters (to fix g++ warnings)
- Remove unnecessary semicolons
- Change type from int to size_t where appropriate
- Fix output error and improve formatting in pointcloud_to_pcd.cpp
- Add some code in concatenate_data.cpp that has been seemingly forgotten
- Other minor things

### DIFF
--- a/pcl_ros/src/pcl_ros/io/concatenate_data.cpp
+++ b/pcl_ros/src/pcl_ros/io/concatenate_data.cpp
@@ -36,7 +36,6 @@
  */
 
 #include <pluginlib/class_list_macros.h>
-#include <pcl/io/io.h>
 #include "pcl_ros/transforms.h"
 #include "pcl_ros/io/concatenate_data.h"
 
@@ -196,7 +195,7 @@ pcl_ros::PointCloudConcatenateDataSynchronizer::subscribe ()
 void
 pcl_ros::PointCloudConcatenateDataSynchronizer::unsubscribe ()
 {
-  for (int d = 0; d < filters_.size (); ++d)
+  for (size_t d = 0; d < filters_.size (); ++d)
   {
     filters_[d]->unsubscribe ();
   }
@@ -272,6 +271,10 @@ pcl_ros::PointCloudConcatenateDataSynchronizer::input (
           {
             pcl_ros::PointCloudConcatenateDataSynchronizer::combineClouds (*out1, *in7, *out2);
             out1 = out2;
+            if (in8 && in8->width * in8->height > 0)
+            {
+              pcl_ros::PointCloudConcatenateDataSynchronizer::combineClouds (*out2, *in8, *out1);
+            }
           } 
         }
       }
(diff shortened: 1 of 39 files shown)


pcl_ros/src/pcl_ros/io/concatenate_data.cpp
```diff
@@ -272,6 +271,10 @@ pcl_ros::PointCloudConcatenateDataSynchronizer::input (
           {
             pcl_ros::PointCloudConcatenateDataSynchronizer::combineClouds (*out1, *in7, *out2);
             out1 = out2;
+            if (in8 && in8->width * in8->height > 0)
```
SteveMacenski, line 274: Can you explain this? `in8` should check if valid and the wdith/height tells you if you have data, in most real applications this should be true so you're adding code for the majority of cases. What's this accomplishing and is there a ticket for it?

SteveMacenski (changes requested):
No issue with the linting portions -- questions on the added code

pcl_ros/src/pcl_ros/io/concatenate_data.cpp
mvieth: Are you asking about the `if`-statement specifically or about all the 4 lines I added there?
Previously the parameter `in8` was unused and not combined like the others. I would consider that behaviour wrong, so I added the 4 lines to fix that. I added the `if`-statement because the other parameters are treated similarly.

SteveMacenski: Im asking about the added cloud itself. You’re changing the general behavior and I want to understand more about the specific issue you or others have seen that this is necessary. 

This is a substantial default behavior change.

mvieth: The issue is that if there are [8 input topics](https://github.com/ros-perception/perception_pcl/blob/bf693eef0b16fc25261c6430b06c9d3091893fe2/pcl_ros/src/pcl_ros/io/concatenate_data.cpp#L174), the last input topic is not considered, i.e. combined with the other clouds, since the parameter `in8` is not used in the callback function `input`. That is why I added [this line](https://github.com/ros-perception/perception_pcl/pull/247/files#diff-5e2bb8ea8ac6f5a7d19fc0fc81a3fe26R276) where `combineClouds` is called with `in8`. And as mentioned, the `if`-statement just checks if the cloud is valid and has points, like it is done for the other inputs as well.

IMO, the fact that the current behaviour ignores `in8` is not what one would expect, and therefore a bug. The proposed change just results in all input clouds being processed equally.

SteveMacenski (approved):

SteveMacenski:
I generally like 2 reviews per PR on high traffic projects. @wjwwood can you glance over it? Pretty small changes

Issue #228, opened 2019-04-05
Bug in transform.hpp?


PetrichSchmoin:
Hello everyone,

is there a bug in setting the `frame_id` of the output cloud in transform.hpp? I always get an empty `frame_id` for the output cloud when I call `transformPointcloud(..)`

In the transform.hpp at the lines [183 to 186](https://github.com/ros-perception/perception_pcl/blob/a8ba2c790db5db181725606649e1fc7c776c1ffd/pcl_ros/include/pcl_ros/impl/transforms.hpp#L183) (and [216 - 219](https://github.com/ros-perception/perception_pcl/blob/a8ba2c790db5db181725606649e1fc7c776c1ffd/pcl_ros/include/pcl_ros/impl/transforms.hpp#L216))
```
  cloud_out.header.frame_id = target_frame;  // <-- here the frame is set
  std_msgs::Header header;
  header.stamp = target_time;
  cloud_out.header = toPCL(header); // <-- and will be overwritten here by an empty frame_id
```

Can anybody confirm this?

My setup:
Ubuntu 16.04
ROS Kinetic 
PCL 1.7

LucidOne:
I updated your post with links to the code, as I understand it.
This looks like a bug, and it should also be an easy unit test to write.
PRs as always are appreciated.

LucidOne:
The test should be something like [this](https://github.com/ros-perception/perception_pcl/blob/melodic-devel/pcl_conversions/test/test_pcl_conversions.cpp)

PetrichSchmoin:
I don't have much time to spend at the moment. Maybe I can try to fix it next month if no one else fixed it until then.

SteveMacenski:
@PetrichSchmoin any update? I would merge that PR pretty quickly

mvieth:
I created a PR. There is no unit test, but the problem itself should be fixed.

PR #249, opened 2019-11-15
Fix problem that frame id is overwritten in cloud transform

mvieth:
Previously, the frame id was empty, now it is the `target_frame`. Also the sequence is set explicitly. Finally, there were some mistakes in the documentation.

See issue #228


SteveMacenski:
Wouldnt it be better to fix toPCL for a header rather than just replace it?

mvieth:
There is nothing wrong with `toPCL(std_msgs::Header)` (at least I don't see anything wrong with it). But the three deleted lines were only to set the target time, which can be done much simpler and with less overhead with `toPCL(ros::Time)`.

SteveMacenski (approved):
Ah yes. Sorry, my mistake.

SteveMacenski:
Lets leave this to soak for a week if anyone has objections.

SteveMacenski:
Merging, thanks!

PR #251, opened 2019-11-19
Split off pcl_ros_filter into separate library for Kinetic

### LINKED ISSUE #16: Filter class should be compiled separately from pcl_ros_filters library

standmit:
Just cherry-pick of #239 to resolve #16 on kinetic branch

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -81,8 +81,7 @@ generate_dynamic_reconfigure_options(
 catkin_package(
   INCLUDE_DIRS include
   LIBRARIES
-    pcl_ros_filters
-    pcl_ros_io
+    pcl_ros_filter
     pcl_ros_tf
   CATKIN_DEPENDS
     dynamic_reconfigure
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Given Kinetic is an old distro in only maintained mode, I’m inclined to not merge this change. I generally only backport important bug patches. 

But if @paulbovbel disagrees, I’ll take his judgement.

paulbovbel:
It's a bug patch, so if the backport looks good, and we have some idea that @standmit has tested it, i'm not against merging and releasing it.

SteveMacenski:
CI failed on a network timeout, restarting

SteveMacenski:
Closing from inactivity, feel free to reopen and fix CI if we want this to be merged.

PR #253, opened 2019-11-26
changing base version to 2.1.0 for first ros2 sync [eloquent]

SteveMacenski:


SteveMacenski:
Need to update 2.0.0 to something else, 2.1.0

PR #256, opened 2019-12-11
Remove redundant source files in src/pcl_ros/filters/features

meyerj:
I assume those files have been added accidentally in a3701bb3df14ba08df641fd6967264b73e15c060 and are actually unreferenced.

They have barely been touched since 2012.


SteveMacenski:
That... looks good to me. Wow can't believe I never noticed that.

SteveMacenski (approved):

PR #258, opened 2019-12-29
Add static message field mapping to Serializer class

taketwo:
Currently, `Serializer` caches computed message field mapping in a [protected field](https://github.com/taketwo/pcl/blob/0a6508ab6c18b5adff3aaf6ef1d2884647f9acba/common/include/pcl/point_cloud.h#L606) of `pcl::PointCloud<T>` class. This commit adds a static field to the Serializer class and stores the mapping there instead.

Motivation: as a part of `boost::shared_ptr -> std::shared_ptr` [migration](https://github.com/PointCloudLibrary/pcl/issues/2792) in PCL we have changed the type of the aforementioned protected field. Later on we were informed that this broke `pcl_ros`. This was unexpected since the field and the associated accessor function are not a public API. After some investigation, we came to the conclusion that `pcl_ros` is probably the only user of that field. With this in mind, we'd like to get rid of it altogether. Therefore, this commit serves two purposes: unbreak pcl_ros for PCL master and get rid of the dependency on non-public API that is going to be deprecated/removed soon.

Implementation note: I've added a mutex to avoid problems if `Serializer` is used simultaneously from several threads for the first time. Not sure if this is needed; is such a situation possible?

### DIFF
--- a/pcl_ros/include/pcl_ros/point_cloud.h
+++ b/pcl_ros/include/pcl_ros/point_cloud.h
@@ -59,29 +59,6 @@ namespace pcl
 
 namespace ros 
 {
-  // In ROS 1.3.1+, we can specialize the functor used to create PointCloud<T> objects
-  // on the subscriber side. This allows us to generate the mapping between message
-  // data and object fields only once and reuse it.
-#if ROS_VERSION_MINIMUM(1, 3, 1)
-  template<typename T>
-  struct DefaultMessageCreator<pcl::PointCloud<T> >
-  {
-    boost::shared_ptr<pcl::MsgFieldMap> mapping_;
-
-    DefaultMessageCreator()
-      : mapping_( boost::make_shared<pcl::MsgFieldMap>() )
-    {
-    }
-    
-    boost::shared_ptr<pcl::PointCloud<T> > operator() ()
-    {
-      boost::shared_ptr<pcl::PointCloud<T> > msg (new pcl::PointCloud<T> ());
-      pcl::detail::getMapping(*msg) = mapping_;
-      return msg;
-    }
-  };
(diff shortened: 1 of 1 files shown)


SteveMacenski:
It looks like your build is failing from the `mapping_` static variable.

taketwo:
Indeed, I forgot to define these variables.

Since this header has no associated `cpp` file, I now moved these static variables inside the method where they are used.

SteveMacenski (approved):
Got it, looks good to me. I'll let this sit for a week or two in case anyone has any objects to raise. 

(Feel free to poke me in this PR if I don't touch base in 2-3 weeks)

taketwo:
Ping.

We ended up reverting that breaking change in PCL and doing a proper deprecation. Thus `pcl_ros` is still compatible with the recently released PCL 1.10.0 and PCL `master`. However, it will generate warnings and moving forward this PR is nevertheless needed.

SteveMacenski:
Thanks for the reminder

mark-goodall:
I've had an issue compiling this from a clean checkout where boost::mutex is not defined. You possibly need to add `#include <boost/thread/mutex.hpp>`, or is this because I'm building against the wrong version of PCL (1.10)?

taketwo:
You are right, this PR introduced a dependency on `boost::mutex`, so the corresponding header should be included. Would you mind sending a patch?

Somehow I did not get compilation errors when I developed this in December. There were some clean-ups/reorganizations of includes in PCL headers recently. Perhaps mutex was indirectly included before, but not anymore, hence the problem.

mark-goodall:
Done https://github.com/ros-perception/perception_pcl/pull/267

PR #264, opened 2020-02-25
Use std::uint* types.

mikepurvis:
Noetic needs this as Focal now has PCL 1.10:

https://packages.ubuntu.com/focal/amd64/libpcl-dev

And it throws a ton of these warnings on any package that includes a pcl_ros or pcl_conversions header:

```
/home/administrator/ros_ws/install/include/pcl_conversions/pcl_conversions.h: In static member function static uint32_t ros::serialization::Serializer<pcl::PCLPointCloud2>::serializedLength(const pcl::PCLPointCloud2&):
/home/administrator/ros_ws/install/include/pcl_conversions/pcl_conversions.h:825:47: error: using uint8_t = uint8_t is deprecated: use std::uint8_t instead of pcl::uint8_t [-Werror=deprecated-declarations]
  825 |         length += m.data.size() * sizeof(pcl::uint8_t);
```

----

This change is *not* safe on Xenial (PCL 1.7), but *is* safe on Bionic (PCL 1.8), where the pcl::int types are already aliased to the `std` versions, just without the deprecation warnings. So it should be fine to go in for either just Noetic or for Noetic + Melodic.

### DIFF
--- a/pcl_conversions/include/pcl_conversions/pcl_conversions.h
+++ b/pcl_conversions/include/pcl_conversions/pcl_conversions.h
@@ -77,19 +77,19 @@ namespace pcl_conversions {
   /** PCLHeader <=> Header **/
 
   inline
-  void fromPCL(const pcl::uint64_t &pcl_stamp, ros::Time &stamp)
+  void fromPCL(const std::uint64_t &pcl_stamp, ros::Time &stamp)
   {
     stamp.fromNSec(pcl_stamp * 1000ull);  // Convert from us to ns
   }
 
   inline
-  void toPCL(const ros::Time &stamp, pcl::uint64_t &pcl_stamp)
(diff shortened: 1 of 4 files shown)


SteveMacenski (approved):

SteveMacenski:
You mind updating that in Dashing as well?

mikepurvis:
Okay, I think I have them all fixed now— a few extras needed prefixing because these packages add functions to `namespace pcl`, where unprefixed `int_t` usages were tripping up the deprecation logic.

I can pick this over to dashing, but it's obviously a separate PR and I'm not set up to test it.

SteveMacenski:
@mikepurvis just to be clear - you're done with modifications or more still required?

mikepurvis:
Both are now good to go in.

SteveMacenski:
Thanks for the help!

PR #265, opened 2020-02-25
Use std::uint* types.

mikepurvis:
ROS2 counterpart to #264.


SteveMacenski (approved):

ruffsl:
@SteveMacenski , does this patch need to be applied to `eloquent-devel` as well? Which is the head ros2 development branch for this repo?

SteveMacenski:
@ruffsl yes, that would be great. Can you submit a PR?

PR #266, opened 2020-03-16
Use std::uint* types. (#265)

ruffsl:
Cherry pick of https://github.com/ros-perception/perception_pcl/pull/265 for eloquent

( @SteveMacenski, anything else that eloquent is out of sync with compared to patches for dashing)


SteveMacenski (approved):

SteveMacenski:
I think that's it from looking at the git history.

PR #267, opened 2020-03-23
Added missing header for boost::mutex

mark-goodall:
ros-perception/perception_pcl#258 Added a mutex to the point cloud serialisation code without adding an include to define boost::mutex. On a clean compile this missing dependency causes the compilation to fail


taketwo (approved):

SteveMacenski:
Retriggering ci

SteveMacenski (approved):
3rd time's the charm on CI

Awesome, thanks

PR #273, opened 2020-05-04
Changes in preparation for PCL 1.11

kunaltyagi:
PCL 1.11 changes it's internal shared pointer from `boost::shared_ptr` to `std::shared_ptr`

This PR addresses the few locations where compiling against PCL 1.11 would fail

PS: Once merged, we can test PCL for the release against melodic and possibly kinetic

PR #274, opened 2020-05-04
Deprecation warning test

kunaltyagi:
Based on https://github.com/ros-perception/perception_pcl/pull/273#issuecomment-623635713

### DIFF
--- a/pcl_ros/include/pcl_ros/pcl_nodelet.h
+++ b/pcl_ros/include/pcl_ros/pcl_nodelet.h
@@ -37,6 +37,8 @@
 
 /**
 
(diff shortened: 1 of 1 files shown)


kunaltyagi:
@SteveMacenski Npr-focal fails on `melodic-devel`

SteveMacenski:
@kunaltyagi I'm pretty sure that's from the upgrade to 1.10 for the Noetic supported PCL version: https://www.ros.org/reps/rep-0003.html#noetic-ninjemys-may-2020-may-2025 and the deprecation warning. 

Fine, did it here https://github.com/ros-perception/perception_pcl/pull/275

PR #277, opened 2020-05-08
Changes in preparation for PCL 1.11

kunaltyagi:
Same patch as #273, generated by:
```
git rebase -i 1459cae  --onto dashing-devel
```
Here 1459cae is the commit id of `melodic-devel~1` (before my patch was merged)

I haven't tested this patch, yet (due to lack of time). Any help regarding the setup for ROS2 would be appreciated. The dockerfile for testing ROS1 melodic PR is located [here](https://github.com/kunaltyagi/pcl/blob/release/.dev/docker/perception_pcl_ros/Dockerfile)

PS: I saw that in the PR for ROS2, boost pointers were replaced by std pointers (but it is unfinished), which would make this PR needless.

### DIFF
--- a/pcl_ros/include/pcl_ros/features/feature.h
+++ b/pcl_ros/include/pcl_ros/features/feature.h
@@ -70,11 +70,11 @@ namespace pcl_ros
       typedef pcl::KdTree<pcl::PointXYZ>::Ptr KdTreePtr;
 
       typedef pcl::PointCloud<pcl::PointXYZ> PointCloudIn;
-      typedef PointCloudIn::Ptr PointCloudInPtr;
-      typedef PointCloudIn::ConstPtr PointCloudInConstPtr;
+      typedef boost::shared_ptr<PointCloudIn> PointCloudInPtr;
+      typedef boost::shared_ptr<const PointCloudIn> PointCloudInConstPtr;
 
-      typedef boost::shared_ptr <std::vector<int> > IndicesPtr;
-      typedef boost::shared_ptr <const std::vector<int> > IndicesConstPtr;
+      typedef pcl::IndicesPtr IndicesPtr;
+      typedef pcl::IndicesConstPtr IndicesConstPtr;
 
       /** \brief Empty constructor. */
       Feature () : /*input_(), indices_(), surface_(), */tree_(), k_(0), search_radius_(0),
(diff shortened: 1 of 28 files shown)


SteveMacenski:
I think you should also have a PR for eloquent-devel, since that will be what rolls into Foxy for the next LTS release.

kunaltyagi:
Could you help me a bit by modifying the original ROS1 dockerfile to work for ROS2?

SteveMacenski:
@kunaltyagi I'm not sure what you're referencing. What dockerfile? 

Travis seems to be configured correctly: https://github.com/ros-perception/perception_pcl/blob/eloquent-devel/.travis.yml

SteveMacenski:
@kunaltyagi do you want this to merge? Its still in draft form. I think maybe this shouldn't be merged since the version of PCL for dashing is 1.8.1. As long as it doesn't break compatibility (which it doesn't appear to) I'm also OK merging as well.

kunaltyagi:
I want to actually test compilation with different PCL versions before merging it. I created the PR for eloquent so that the work for foxy isn't delayed due to my time constraints. Hence the delay.

I'm reasonably sure that the patch should work with PCL 1.8 through 1.11 but I haven't tested that hypothesis.

AswinkarthikeyenAK:
Hi, 
while doing a catkin_make, I am getting the following error,
`/home/catkin_ws/src/perception_pcl/pcl_ros/include/pcl_ros/surface/moving_least_squares.h:69:18: error: ‘KdTree’ in namespace ‘pcl’ does not name a template type
     typedef pcl::KdTree<PointIn> KdTree;
                  ^~~~~~
/home/catkin_ws/src/perception_pcl/pcl_ros/include/pcl_ros/surface/moving_least_squares.h:70:18: error: ‘KdTree’ in namespace ‘pcl’ does not name a template type
     typedef pcl::KdTree<PointIn>::Ptr KdTreePtr;
                  ^~~~~~
/home/catkin_ws/src/perception_pcl/pcl_ros/include/pcl_ros/surface/moving_least_squares.h:77:7: error: ‘KdTreePtr’ does not name a type; did you mean ‘IndicesPtr’?
       KdTreePtr tree_;
       ^~~~~~~~~
       IndicesPtr
`
I have installed the pcl version 1.11.1
Any suggestion regarding this is really appreciated. Thanks in advance

kunaltyagi:
Closing PR because Dashing is EOL.

Regarding the error, it seems like the mls doesn't include kdtree header. I suggest you create a new issue/PR for that

PR #279, opened 2020-05-12
Changes in preparation for PCL 1.11

kunaltyagi:
As recommended by https://github.com/ros-perception/perception_pcl/pull/277#issuecomment-625589015.

Any patches found while testing #277 will be applied uniformly to the other 2 PRs


SteveMacenski (approved):

SteveMacenski:
I was about to say - that's strange that N/M ran, but I see a base change. 

LGTM

PR #285, opened 2020-06-26
Add workaround for deprecated filter functions

mvieth:
See also:
https://github.com/PointCloudLibrary/pcl/commit/f26c6fcc24386c29213b62423710c8927cbe1d21
https://github.com/PointCloudLibrary/pcl/commit/630627b2accad395f80bc6518f8dcd94a1f9ad31

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -4,10 +4,10 @@ project(pcl_ros)
 add_compile_options(-std=c++14)
 
 ## Find system dependencies
+find_package(PCL REQUIRED COMPONENTS core features filters io segmentation surface)
 find_package(cmake_modules REQUIRED)
 find_package(Boost REQUIRED COMPONENTS system filesystem thread)
 find_package(Eigen3 REQUIRED)
-find_package(PCL REQUIRED COMPONENTS core features filters io segmentation surface)
 
 if(NOT "${PCL_LIBRARIES}" STREQUAL "")
   # For debian: https://github.com/ros-perception/perception_pcl/issues/139
--- a/pcl_ros/src/pcl_ros/filters/passthrough.cpp
+++ b/pcl_ros/src/pcl_ros/filters/passthrough.cpp
@@ -95,11 +95,19 @@ pcl_ros::PassThrough::config_callback (pcl_ros::FilterConfig &config, uint32_t /
   }
 
   // Check the current value for the negative flag
+#if PCL_VERSION_COMPARE(<, 1, 10, 0)
   if (impl_.getFilterLimitsNegative () != config.filter_limit_negative)
+#else
+  if (impl_.getNegative () != config.filter_limit_negative)
+#endif
   {
     NODELET_DEBUG ("[%s::config_callback] Setting the filter negative flag to: %s.", getName ().c_str (), config.filter_limit_negative ? "true" : "false");
     // Call the virtual method in the child
+#if PCL_VERSION_COMPARE(<, 1, 10, 0)
     impl_.setFilterLimitsNegative (config.filter_limit_negative);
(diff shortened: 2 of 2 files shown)


SteveMacenski:
I'm OK with this once CI passes

mvieth:
Ping @SteveMacenski The reasons why the checks fail seem unrelated

SteveMacenski:
Toggling to rerun, lets see what it says now. The noetic failure was caused by `Eigen` rather than `Eigen3` rosdep and CMakelists entries. The Melodic looks like a fluke but lets see. 

Ping me when it finishes and we can take a look again. If you're so motivated to change Eigen -> Eigen3 that would be great.

mvieth:
@SteveMacenski ok, melodic passes now. The Eigen thing comes from pcl itself.

SteveMacenski:
I think the Eigen thing is because the cmakelists / package.xml of `pcl_ros` points at Eigen and not `Eigen3`. I've experienced this issue in other packages as well, its a really easy "add a 3" situation anywhere Eigen exists. I'd appreciate if you were to add that, but if not, I can do it at a larger date.

SteveMacenski (approved):

mvieth:
I am quite sure that the Eigen issue has to be fixed in pcl itself, not here in perception_pcl. Everywhere here, Eigen is already called Eigen3, except in the package.xml, but even with that changed the cmake warning still pops up. The log says:
```
/usr/lib/x86_64-linux-gnu/cmake/pcl/PCLConfig.cmake:152 (find_package)
/usr/lib/x86_64-linux-gnu/cmake/pcl/PCLConfig.cmake:316 (find_eigen)
/usr/lib/x86_64-linux-gnu/cmake/pcl/PCLConfig.cmake:551 (find_external_library)
CMakeLists.txt:10 (find_package)
```
PCL has a FindEigen.cmake that actually finds Eigen3, and there is a FindEigen.cmake in the ros distribution under share/cmake_modules that is deprecated. I guess these two get mixed up and that causes the problem. I might create a PR for this issue in the pcl repo.

mvieth:
After thinking a bit more about the Eigen issue, it seems there is a really simple solution after all. The warning disappears just by moving the `find_package(PCL)` above the `find_package(cmake_modules)` in the CMakeLists of pcl_ros. Maybe the dependence on cmake_modules could be even completely removed, it seems to be not needed (any more).

SteveMacenski:
@mvieth please make that change then

mvieth:
@SteveMacenski done. No cmake warnings anymore.

SteveMacenski:
Still warnings 

```
05:15:46 /tmp/ws/src/perception_pcl/pcl_ros/src/pcl_ros/surface/moving_least_squares.cpp:215:48: warning: ‘void pcl::MovingLeastSquares<PointInT, PointOutT>::setPolynomialFit(bool) [with PointInT = pcl::PointXYZ; PointOutT = pcl::PointNormal]’ is deprecated: use setPolynomialOrder() instead [-Wdeprecated-declarations]
05:15:46   215 |     impl_.setPolynomialFit (use_polynomial_fit_);
05:15:46       |                                                ^
05:15:46 In file included from /tmp/ws/src/perception_pcl/pcl_ros/include/pcl_ros/surface/moving_least_squares.h:44,
05:15:46                  from /tmp/ws/src/perception_pcl/pcl_ros/src/pcl_ros/surface/moving_least_squares.cpp:39:
05:15:46 /usr/include/pcl-1.10/pcl/surface/mls.h:361:7: note: declared here
05:15:46   361 |       setPolynomialFit (bool polynomial_fit)
05:15:46       |       ^~~~~~~~~~~~~~~~
```

but not your problem in this PR. I'm good with this

PR #290, opened 2020-07-24
Port transforms library to ROS2

theseankelly:
- Port the transforms library to ROS2
- Update CMakeLists
- Update package.xml
- Enable the package

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -146,6 +146,9 @@ target_link_libraries(pcl_ros_tf ${PCL_LIBRARIES})
 #############
 
 if(BUILD_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+
   find_package(ament_cmake_gtest REQUIRED)
   #add_rostest_gtest(test_tf_message_filter_pcl tests/test_tf_message_filter_pcl.launch src/test/test_tf_message_filter_pcl.cpp)
   #target_link_libraries(test_tf_message_filter_pcl ${catkin_LIBRARIES} ${GTEST_LIBRARIES})
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -1,245 +1,89 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.5)
 project(pcl_ros)
 
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
 ## Find system dependencies
-find_package(cmake_modules REQUIRED)
-find_package(Boost REQUIRED COMPONENTS system filesystem thread)
 find_package(Eigen3 REQUIRED)
-find_package(PCL REQUIRED COMPONENTS core features filters io segmentation surface)
+find_package(PCL REQUIRED QUIET COMPONENTS core features filters io segmentation surface)
 
-if(NOT "${PCL_LIBRARIES}" STREQUAL "")
-  # For debian: https://github.com/ros-perception/perception_pcl/issues/139
-  list(REMOVE_ITEM PCL_LIBRARIES
-    "vtkGUISupportQt"
-    "vtkGUISupportQtOpenGL"
-    "vtkGUISupportQtSQL"
-    "vtkGUISupportQtWebkit"
-    "vtkViewsQt"
-    "vtkRenderingQt")
-endif()
+## Find ROS package dependencies
+find_package(ament_cmake_ros REQUIRED)
+find_package(pcl_conversions REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(sensor_msgs REQUIRED)
+find_package(tf2 REQUIRED)
+find_package(tf2_geometry_msgs REQUIRED)
+find_package(tf2_ros REQUIRED)
 
-# There is a bug in the Ubuntu Artful (17.10) version of the VTK package,
-# where it includes /usr/include/*-linux-gnu/freetype2 in the include
-# directories (which doesn't exist).  This filters down to the PCL_INCLUDE_DIRS,
-# and causes downstream projects trying to use these libraries to fail to
-# configure properly.  Here we remove those bogus entries so that downstream
-# consumers of this package succeed.
-if(NOT "${PCL_INCLUDE_DIRS}" STREQUAL "")
-  foreach(item ${PCL_INCLUDE_DIRS})
-    string(REGEX MATCH "/usr/include/.*-linux-gnu/freetype2" item ${item})
-    if(item)
-      list(REMOVE_ITEM PCL_INCLUDE_DIRS ${item})
-    endif()
-  endforeach()
-endif()
-
-## Find catkin packages
-find_package(catkin REQUIRED COMPONENTS
-  dynamic_reconfigure
-  message_filters
-  nodelet
-  nodelet_topic_tools
+ament_export_dependencies(
   pcl_conversions
-  pcl_msgs
-  pluginlib
-  rosbag
-  rosconsole
-  roscpp
-  roslib
+  rclcpp
   sensor_msgs
-  std_msgs
-  tf
-  tf2_eigen
-)
-
-## Add include directories
-include_directories(include
-  ${Boost_INCLUDE_DIRS}
-  ${catkin_INCLUDE_DIRS}
-  ${EIGEN3_INCLUDE_DIRS}
-  ${PCL_INCLUDE_DIRS}
-)
-
-## Generate dynamic_reconfigure
-generate_dynamic_reconfigure_options(
-  cfg/EuclideanClusterExtraction.cfg
-  cfg/ExtractIndices.cfg
-  cfg/ExtractPolygonalPrismData.cfg
-  cfg/Feature.cfg
-  cfg/Filter.cfg
-  cfg/MLS.cfg
-  cfg/SACSegmentation.cfg
-  cfg/SACSegmentationFromNormals.cfg
-  cfg/SegmentDifferences.cfg
-  cfg/StatisticalOutlierRemoval.cfg
-  cfg/RadiusOutlierRemoval.cfg
-  cfg/VoxelGrid.cfg
-  cfg/CropBox.cfg
-)
-
-## Declare the catkin package
-catkin_package(
-  INCLUDE_DIRS include
-  LIBRARIES
-    pcl_ros_filters
-    pcl_ros_io
-    pcl_ros_tf
-  CATKIN_DEPENDS
-    dynamic_reconfigure
-    message_filters
-    nodelet
-    nodelet_topic_tools
-    pcl_conversions
-    pcl_msgs
-    rosbag
-    roscpp
-    sensor_msgs
-    std_msgs
-    tf
-  DEPENDS
-    Boost
-    EIGEN3
-    PCL
+  tf2
+  tf2_geometry_msgs
+  tf2_ros
+  EIGEN3
+  PCL
 )
 
 ## Declare the pcl_ros_tf library
 add_library(pcl_ros_tf src/transforms.cpp)
-target_link_libraries(pcl_ros_tf ${Boost_LIBRARIES} ${catkin_LIBRARIES} ${PCL_LIBRARIES})
-add_dependencies(pcl_ros_tf ${catkin_EXPORTED_TARGETS})
-
-## Nodelets
-
-## Declare the pcl_ros_io library
-add_library(pcl_ros_io
-  src/pcl_ros/io/bag_io.cpp
-  src/pcl_ros/io/concatenate_data.cpp
-  src/pcl_ros/io/concatenate_fields.cpp
-  src/pcl_ros/io/io.cpp
-  src/pcl_ros/io/pcd_io.cpp
+ament_target_dependencies(pcl_ros_tf
+  "pcl_conversions"
+  "rclcpp"
+  "sensor_msgs"
+  "tf2"
+  "tf2_geometry_msgs"
+  "tf2_ros"
 )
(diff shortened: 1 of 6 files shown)


pcl_ros/include/pcl_ros/impl/transforms.hpp
```diff
@@ -92,30 +106,33 @@ template <typename PointT> bool
...
+  catch (tf2::ExtrapolationException &e)
   {
-    ROS_ERROR ("%s", e.what ());
+    RCLCPP_ERROR (rclcpp::get_logger("pcl_ros"), "%s", e.what ());
     return (false);
   }
 
+  tf2::Transform transform;
+  tf2::convert (transform_msg.transform, transform);
+
```
theseankelly: This conversion step is a candidate for a helper function, but some feedback is needed.

This is more about the tf->tf2 port than the ROS2 port. Given that the `lookupTransform` API in tf2 returns `geometry_msgs::msg::TransformStamped` instead of `tf2::Transform` directly, I wonder if this calls for the creation of a new overload of the `transformPointCloud[WithNormals]` APIs like so:

```
  template <typename PointT> void 
  transformPointCloudWithNormals (const pcl::PointCloud <PointT> &cloud_in, 
                                  pcl::PointCloud <PointT> &cloud_out, 
                                  const geometry_msgs::msg::Transform &transform);
```

Which would provide this conversion. Obviously this pulls `geometry_msgs` into the public interface.

Additionally, would the existing API which takes a `tf2::Transform` be considered deprecated?

pcl_ros/src/transforms.cpp
```diff
@@ -55,22 +69,25 @@ transformPointCloud (const std::string &target_frame, const sensor_msgs::PointCl
...
+  catch (tf2::ExtrapolationException &e)
   {
-    ROS_ERROR ("%s", e.what ());
+    RCLCPP_ERROR (rclcpp::get_logger("pcl_ros"), "%s", e.what ());
     return (false);
   }
 
+  tf2::Transform transform;
+  tf2::convert (transform_msg.transform, transform);
+
```
theseankelly: Same discussion points here as in transforms.hpp -- should we provide an overload of `transformAsMatrix` which accepts a `const geometry_msgs::msg::Transform&` instead of a `tf2::Transform`?

theseankelly:
The most basic unit tests which exercise this code also require the filters package to be ported.

In order to keep this initial PR as standalone as possible, I've manually tested the API with my 3d camera. I'll be working on the filters package in order to enable the unit tests next.

pcl_ros/CMakeLists.txt
```diff
@@ -1,245 +1,89 @@
...
+## Find ROS package dependencies
+find_package(ament_cmake_ros REQUIRED)
+find_package(pcl_conversions REQUIRED)
+find_package(rclcpp REQUIRED)
+find_package(sensor_msgs REQUIRED)
+find_package(tf2 REQUIRED)
+find_package(tf2_geometry_msgs REQUIRED)
+find_package(tf2_ros REQUIRED)
+
+ament_export_dependencies(
```
SteveMacenski: The ament_* macros are typically at the end of the files

SteveMacenski: Usually we try to: `set(depedencies ...)` and then use `${dependencies}` in the export / linking so that we don't have multiple lists to edit to add a new dependency

pcl_ros/CMakeLists.txt
```diff
@@ -1,245 +1,89 @@
...
+  RUNTIME DESTINATION bin
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
 )
 
 install(DIRECTORY plugins samples
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION})
+  DESTINATION share/${PROJECT_NAME})
+
+ament_export_targets(export_pcl_ros HAS_LIBRARY_TARGET)
```
SteveMacenski, line 191: Around here is where that ament_export_dependencies should go

pcl_ros/CMakeLists.txt
```diff
@@ -1,245 +1,89 @@
...
-  src/pcl_ros/filters/statistical_outlier_removal.cpp
-  src/pcl_ros/filters/voxel_grid.cpp
-  src/pcl_ros/filters/crop_box.cpp
-)
-target_link_libraries(pcl_ros_filters pcl_ros_tf ${Boost_LIBRARIES} ${catkin_LIBRARIES} ${PCL_LIBRARIES})
-add_dependencies(pcl_ros_filters ${PROJECT_NAME}_gencfg)
-class_loader_hide_library_symbols(pcl_ros_filters)
-
-## Declare the pcl_ros_segmentation library
-add_library (pcl_ros_segmentation
```
SteveMacenski, line 160: Please don't delete this block, just comment it out. Its good to have a base to start from for others to port

pcl_ros/CMakeLists.txt
```diff
@@ -1,245 +1,89 @@
...
+    $<INSTALL_INTERFACE:include>
+    ${PCL_INCLUDE_DIRS})
+target_link_libraries(pcl_ros_tf ${PCL_LIBRARIES})
+ament_export_libraries(pcl_ros_tf)
 
 #############
 ## Testing ##
 #############
 
-if(CATKIN_ENABLE_TESTING)
```
SteveMacenski, line 211: Are none of these tests enabled by the pcl_ros_tf that you ported?

pcl_ros/CMakeLists.txt
```diff
@@ -1,245 +1,89 @@
...
-  target_link_libraries(test_tf_message_filter_pcl ${catkin_LIBRARIES} ${GTEST_LIBRARIES})
-  add_rostest(samples/pcl_ros/features/sample_normal_3d.launch ARGS gui:=false)
-  add_rostest(samples/pcl_ros/filters/sample_statistical_outlier_removal.launch ARGS gui:=false)
-  add_rostest(samples/pcl_ros/filters/sample_voxel_grid.launch ARGS gui:=false)
-  add_rostest(samples/pcl_ros/segmentation/sample_extract_clusters.launch ARGS gui:=false)
-  add_rostest(samples/pcl_ros/surface/sample_convex_hull.launch ARGS gui:=false)
-endif(CATKIN_ENABLE_TESTING)
+if(BUILD_TESTING)
+  find_package(ament_cmake_gtest REQUIRED)
+  # TODO
```
SteveMacenski: Add the auto ament linters please

SteveMacenski (changes requested):

pcl_ros/include/pcl_ros/impl/transforms.hpp
SteveMacenski: Sure, makes sense to support both.

pcl_ros/src/transforms.cpp
SteveMacenski: Or change to just geometry_msgs if TF2 doesn't give back out TF2 types (which... seems odd but OK)

theseankelly: Agreed it's quite odd since TF2 claims to be ROS independent. But it seems to be the case unless I'm missing something...I spent quite a while searching. Here's the API in the tf2 core library:

https://github.com/ros/geometry2/blob/noetic-devel/tf2/include/tf2/buffer_core.h

pcl_ros/CMakeLists.txt
theseankelly: I opted against the global ${dependencies} because each sub-component is going to depend and link against very different things. It felt wrong to just broadly link against everything for each component (from the cmake/export perspective). But I guess it doesn't really matter.

pcl_ros/CMakeLists.txt
theseankelly: Correct. I started down enabling the tests under `pcl_ros\src\test\test_tf_message_filter_pcl.cpp` but determined that it depends on the pcl_ros_filters module, and decided to take it one step at a time.

However on a second look, it appears I was wrong and it only depends on the *tf2* filters package. It does depend on `pcl_ros\point_cloud.h` though, which has not been ported by this PR.

I'm leaning towards expanding the scope of this PR a bit to get `point_cloud.h` and the first round of unit tests ported as well -- cool? Or would you rather keep the PR smaller?

pcl_ros/CMakeLists.txt
SteveMacenski: You could also make different dependency variables for different sets of components, but hardcoding `N` times `M` dependencies is going to be error prone to deal with if we add more.

pcl_ros/CMakeLists.txt
SteveMacenski: Sounds good - if its just 1 header file extra, that's OK

theseankelly: Ah, I've refreshed my memory as to where this started to turn into an onion. The tests make use of `pcl_ros/point_cloud.h` which plugs `pcl::PointCloud<T>` into the message serialization/adaptation feature of ROS1. Seems like this isn't yet implemented in ROS2 and the comments on [this thread](https://answers.ros.org/question/356542/ros2-message-serialization-adapting-types/) suggest it might not be a good idea moving forward anyway.

Thoughts? I could modify the tests to use regular ROS message types, but that would kind of fork away from the current implementation.

theseankelly: Turns out I'm just standing on the shoulders of giants --  https://discourse.ros.org/t/ros2-type-masquerading-status-and-alternatives/9461

SteveMacenski: What's the question exactly? I agree that type masquerading isn't available in ROS2 (yet). Does that stop this port or just this file needs to be commented out or something?

theseankelly: Sorry my morning was rushed; should have been more verbose.

The good news is, as far as I can tell, the only component _within_ `pcl_ros` itself which actually uses the type-masquerading header is the `pcl_ros\src\test\test_tf_message_filter_pcl.cpp` which is the test file I've proposed enabling in this thread. So the port can go on, but downstream consumers of pcl_ros which rely on the type masquerading stuff obviously won't work. I don't have any context to tell how critical of a feature this is in the `pcl_ros` library.

As I see it, the options are:

1. Skip porting the test for now until the masquerading is available
2. Port the test now with a bit of rewriting to use `sensor_msgs::PointCloud2` messages instead of `pcl::PointCloud<T>` directly (and down the road change it back to get coverage on the masquerading)
3. Go start shaving the type masquerading yak now

My vote is option 2 to get some automated coverage and keep the `pcl_ros` port on track. Though type masquerading would have widespread benefit it seems.

SteveMacenski: Could you just change the type from the pcl::Pointcloud to the sensor_msgs::Pointcloud2 to avoid type masquerading? Obviously the test will be slower, but that's OK given the state of things. 

Its pretty important for performance there's no conversions from PC2 -> pcl -> PC2 to publish again, but for now we can only do what we can do. PCL now has this `PCLPointCloud2` thing which my understanding is doesn't have to copy around and instead just reformats the data for a limited subset of the filters but works with most here (or slightly more efficient).

SteveMacenski: We can also file a ticket to re-add type masquerading when available, but that might be a ways down the road from the last few discussions I've had with open robotics about this (its hard, time consuming, and so far no one funding it).

SteveMacenski (approved):

pcl_ros/CMakeLists.txt
theseankelly: I've added them, but is your intent for me to make the linter clean as well?

In general when doing a port like this I like to keep the changes as minimal as possible so we don't overload the ROS->ROS2 diffs with a whole bunch of formatting and whitespace changes. Perhaps making the linter clean is a final step in the port once everything else is working?

SteveMacenski:
You happy with this when CI passes to merge?

pcl_ros/CMakeLists.txt
theseankelly: So after starting down this road, I realized these tests aren't actually testing anything other than the type masquerading itself. It's using the `tf2` library as a vehicle, but it is NOT actually testing the `pcl_ros` transforms library.

With that in mind I've held off on porting these tests.

Unfortunately there are no unit tests specifically for the transforms library, but they're sufficiently simple wrappers around PCL and I've gone through and manually tested them. We can file a ticket to add unit testing as an effort outside of the porting if needed.

theseankelly:
I'm good with this if you are (pending my comments about the linter and the unit tests)

SteveMacenski:
OK - you have a bunch of linter problems (see CI)

For the uncrustify ones you can automatically fix with `ament_uncrustify --reformat` which should do about 70% of it. The rest with `ament_cpplint` will need to be done manually

theseankelly:
Yeah I was aware of that, per my comments above -- I don't feel great about combining a port with a large code formatting change; will make it quite hard for folks to follow the actual porting change. But I guess keeping the linters clean from the start is a reasonable policy and we'll just get the big reformat out of the way along with the (relatively) low delta transforms change.

SteveMacenski:
Plus I read over all the code pre-reformat and I'm OK with everything, so I think its safe. Most of them are white lines and lines too long anyhow

theseankelly:
The changes to get linting clean will be quite broad and I don't feel good about mixing formatting changes with porting/code changes -- I think it's important to keep a clean and readable change history. Let's get the reformatting stuff done first before any introducing any functionality changes.

I'm going to work on another PR with the goal of porting the package.xml, the CMakeLists, and clean linters on all existing code. There will be no code changes other than what's required to get the ROS2 linters passing cleanly. (The actual targets will all be disabled in CMakeLists.txt so the build is effectively a no-op).

Once that step is done, I'll re-do the transforms changes on top of the reformatted codebase; which will probably look different enough to merit a whole new PR.

SteveMacenski:
OK, I'd recommend at minimum 2 PRs. One with `ament_uncrustify --reformat` that's 100% done automatically so I can just merge without a deep review. Then the remaining in 1 or more PRs depending on size and your own interest

theseankelly:
Good idea about ament_uncrustify -- I'll do exactly that.

theseankelly:
Hey @SteveMacenski for uncrustify and cpplint it picks up .h files as C files. Do you prefer I rename all .h files to .hpp or set the appropriate cmake variables to make the linters assume C++?

SteveMacenski:
To hpp is best for ROS 2 styling

PR #292, opened 2020-07-30
Add missing include for new PCL versions

mvieth:
After this change https://github.com/PointCloudLibrary/pcl/commit/6df3e602a72ea16657f901c9a6911d95b263ba08#diff-24324a084e511502d7f34054ec31b353L50, an explicit include for KdTree is needed.
It was formerly a transitive include via pcl/surface/mls.h - pcl/search/pcl_search.h - pcl/search/kdtree.h - pcl/kdtree/kdtree_flann.h.


SteveMacenski:
Can you submit this for Foxy too?

SteveMacenski (approved):

PR #298, opened 2020-08-09
Make ament_cpplint pass

theseankelly:
Collection of hand-made changes to make ament_cpplint pass consisting of:
- whitespace between comments
- line length
- header ordering
- include guard formats
- remove a couple `using namespace std;`
- using c++ casts instead of c-style casts


theseankelly:
@SteveMacenski FYI I am running the full suite of linters locally before these stand-alone PRs to ensure I haven't regressed the others (in this case, uncrustify).

PR #300, opened 2020-08-10
Reformat license and copyright text for ament_copyright

theseankelly:
This is only a reformat so that ament_copyright picks up the text
correctly. There are no changes made to the license text or
copyright owner/year

### DIFF
--- a/pcl_ros/cfg/SACSegmentation_common.py
+++ b/pcl_ros/cfg/SACSegmentation_common.py
@@ -1,5 +1,34 @@
 #! /usr/bin/env python
 
+# Copyright 2009 Willow Garage, Inc
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#    * Redistributions of source code must retain the above copyright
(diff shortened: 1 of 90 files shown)


theseankelly:
@SteveMacenski this is the last of the one-off linter changes -- I'll rebase the transforms change (which also enables the package) next which will also take care cleaning `ament_xmllint` and `ament_lint_cmake`

SteveMacenski:
So I'd be OK merging this, but this might be one worth just disabling that particular test

theseankelly:
I kind of agree. I'll add the opt-out directive to the cmake file when I enable linting instead?

SteveMacenski:
Yeah, that would work. I'm mostly concerned about cpplint, uncrustify, flake8 and cppcheck. Those are the "core code" ones that make the styling consistent. The copyright headers if missing always brings up a bunch of issues about who is the copyright holder. We could be opening ourselves to legal challenges if we mis-attribute the new headers

theseankelly:
Great points, thanks @SteveMacenski -- closing request and stay tuned for some changes with real functionality ;)

PR #301, opened 2020-08-11
Port transforms library to ROS2

theseankelly:
- Port the transforms library to ROS2
- Update CMakeLists
- Update package.xml
- Enable the package

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -34,16 +34,16 @@ set(dependencies
   PCL
 )
 
+include_directories(
+  include
+  ${PCL_INCLUDE_DIRS}
+)
+
 ## Declare the pcl_ros_tf library
 add_library(pcl_ros_tf src/transforms.cpp)
 ament_target_dependencies(pcl_ros_tf
   ${dependencies}
 )
-target_include_directories(pcl_ros_tf
-  PUBLIC
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
-    $<INSTALL_INTERFACE:include>
-    ${PCL_INCLUDE_DIRS})
 target_link_libraries(pcl_ros_tf ${PCL_LIBRARIES})
 
 ### Nodelets
@@ -163,8 +163,10 @@ if(BUILD_TESTING)
 endif()
 
 
-install(DIRECTORY include/
-  DESTINATION include)
+install(
+  DIRECTORY include/
+  DESTINATION include
+)
 
 install(
   TARGETS
@@ -183,6 +185,7 @@ install(
   RUNTIME DESTINATION bin
   ARCHIVE DESTINATION lib
   LIBRARY DESTINATION lib
+  INCLUDES DESTINATION include
 )
 
 install(DIRECTORY plugins samples
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -1,245 +1,195 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.5)
 project(pcl_ros)
 
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
 ## Find system dependencies
-find_package(cmake_modules REQUIRED)
-find_package(Boost REQUIRED COMPONENTS system filesystem thread)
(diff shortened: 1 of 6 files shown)


pcl_ros/CMakeLists.txt
```diff
@@ -1,245 +1,195 @@
...
-## Downloads
-
-catkin_download(table_scene_lms400.pcd http://download.ros.org/data/pcl/table_scene_lms400.pcd
-  DESTINATION ${CMAKE_CURRENT_SOURCE_DIR}/samples/data
-  MD5 546b5b4822fb1de21b0cf83d41ad6683
-)
-add_custom_target(download ALL DEPENDS download_extra_data)
+target_include_directories(pcl_ros_tf
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
```
SteveMacenski: Curious what's with this, just doing `include/` should work OK, what's the rest of this for?

pcl_ros/package.xml
```diff
@@ -10,10 +11,8 @@
 
   </description>
 
-  <author>Open Perception</author>
-  <author email="julius@kammerl.de">Julius Kammerl</author>
-  <author email="william@osrfoundation.org">William Woodall</author>
-
+  <maintainer email="julius@kammerl.de">Julius Kammerl</maintainer>
```
SteveMacenski: I don't think either are maintainers, I'd just leave them as authors

pcl_ros/package.xml
```diff
@@ -24,35 +23,23 @@
...
+  <depend>tf2_geometry_msgs</depend>
+  <depend>tf2_ros</depend>
 
-  <test_depend>rostest</test_depend>
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
+  <test_depend>ament_cmake_gtest</test_depend>
 
   <export>
-    <nodelet plugin="${prefix}/plugins/nodelet/libpcl_ros_features.xml"/>
```
SteveMacenski, line 51: Might be good to comment these out so they can be added back later easier

SteveMacenski (commented):
LGTM, just some small stuff

pcl_ros/package.xml
theseankelly: ament_xmllint complains about the author tag being nonexistant

pcl_ros/CMakeLists.txt
theseankelly: It's the recommended way to do modern CMake, at least as far as I'm aware. Deals with target exporting/importing stuff.

Maybe not a concern for ROS packages, am totally fine with old-style too:
```
target_include_directories(pcl_ros_tf include)
```

pcl_ros/package.xml
theseankelly: Agreed, will do. I originally decided to take the approachj 'start clean, only add stuff when needed' instead of keeping old ROS1 cruft around, but since I've added all the ROS1 targets back to the CMakeLists, I should also keep this around but commented.

pcl_ros/package.xml
theseankelly: Probably good to leave ament_xmllint enabled.

Can do either:
1. leave them as maintainers
2. Add comment above the maintainers line to specify the original authors

Either way I was a little surprised to see that the author tag isn't part of the schema, maybe I'm doing something very wrong??

theseankelly: Sorry for all the comments, I'm live streaming this thing apparently.

Looks like the author tag is listed as an optional element in schema v3. And so this is probably actually an ament_xmllint bug -- let me do some additional research to confirm and if so we can disable the xmllinting temporarily and leave the author tags.

https://www.ros.org/reps/rep-0149.html#package-format-3

pcl_ros/package.xml
theseankelly: done

pcl_ros/package.xml
theseankelly: PEBKAC -- it was an ordering issue. Author tag must come after maintainer/license/url etc.

I've changed them back to authors and moved the tags down so ament_xmllint passes.

pcl_ros/CMakeLists.txt
theseankelly: done

SteveMacenski (approved):

mikeferguson:
@theseankelly thanks for this - just used it in porting a downstream package.

SteveMacenski:
@theseankelly I have another person that would like to help in porting pcl_ros to ROS2, can you join the navigation2 slack channel with this link (https://join.slack.com/t/navigation2/shared_invite/zt-gnsw70li-tUsWwB1C9h3FLIabNVvB~A) so we can chat so we don't have redundant efforts?

themightyoarfish:
Why were so many executables (such as `bag_to_pcd`) removed by this PR?

PR #302, opened 2020-08-15
Deprecate use_polynomial_fit

mvieth:
The pcl function setPolynomialFit is deprecated, setPolynomialOrder should be used instead

### DIFF
--- a/pcl_ros/src/pcl_ros/surface/moving_least_squares.cpp
+++ b/pcl_ros/src/pcl_ros/surface/moving_least_squares.cpp
@@ -212,7 +212,22 @@ pcl_ros::MovingLeastSquares::config_callback (MLSConfig &config, uint32_t /*leve
   {
     use_polynomial_fit_ = config.use_polynomial_fit;
     NODELET_DEBUG ("[config_callback] Setting the use_polynomial_fit flag to: %d.", use_polynomial_fit_);
+#if PCL_VERSION_COMPARE(<, 1, 9, 0)
     impl_.setPolynomialFit (use_polynomial_fit_);
+#else
+    if (use_polynomial_fit_)
+    {
+      NODELET_WARN ("[config_callback] use_polynomial_fit is deprecated, use polynomial_order instead!");
+      if (impl_.getPolynomialOrder () < 2)
+      {
+        impl_.setPolynomialOrder (2);
+      }
+    }
+    else
+    {
+      impl_.setPolynomialOrder (0);
+    }
+#endif
   }
   if (polynomial_order_ != config.polynomial_order)
   {


SteveMacenski:
There appears to be some debug code here now (`true`), is this ready for review?

mvieth:
> There appears to be some debug code here now (`true`), is this ready for review?

Yes, that's intentional. `true` is the default value. If someone sets it to false, they get the error message.

SteveMacenski:
You can't hard code that. That's dead code and removing a configuration. If you want to remove it, remove it and use the new API.

mvieth:
Do you mean removing it from [MLS.cfg](https://github.com/ros-perception/perception_pcl/blob/melodic-devel/pcl_ros/cfg/MLS.cfg)? I assumed that could break things, so I intended to "deprecate" that parameter by checking if it was set (not `true`) and warn the user. The function `setPolynomialOrder`, that should be used instead of `setPolynomialFit`, can already be accessed by the parameter `polynomial_order`.

SteveMacenski:
Keep the parameter and do preprocessor expressions for the version of PCL. For melodic, that might still be valid for the version of PCL it uses. For Noetic, it might be gone / deprecation warning with that version of PCL and there should be a warning and then use the moved flag. I wouldn't modify the cfg file.

mvieth:
I do not think that a preprocessor solution, like [I used for that other PR](https://github.com/ros-perception/perception_pcl/pull/285/files), is appropriate here. The [replacement function](https://github.com/PointCloudLibrary/pcl/blob/master/surface/include/pcl/surface/mls.h#L349), `setPolynomialOrder`, has been around for about 9 years, so I do not see any reason why it should not be used in all ROS versions, even if `setPolynomialFit` is not deprecated yet in older PCL versions.

SteveMacenski:
> The replacement function, setPolynomialOrder, has been around for about 9 years, so I do not see any reason why it should not be used in all ROS versions, even if setPolynomialFit is not deprecated yet in older PCL versions.

I was unaware of that. OK. Do these both do the same thing? If so, I think this is the best solution:
- Keep warning, and add to the warning that its going to use `setPolynomialOrder` instead.
- Set using `setPolynomialOrder` in the `setPolynomialFit` setting

This will help in a couple of ways:
- Melodic and Noetic have already been released, changing the ABI is bad, this is the least bad way to do it
- Users still get a warning that something's wrong to switch to the modern way

If they are fundamentally different, then maybe we have the warning be an error and do the same thing. We can't "not" support it anymore, its already been released and this is the last ROS1 distribution so we're a little SOL on this one.

mvieth:
I changed the commit, let me know what you think

pcl_ros/src/pcl_ros/surface/moving_least_squares.cpp
```diff
@@ -212,7 +212,22 @@ pcl_ros::MovingLeastSquares::config_callback (MLSConfig &config, uint32_t /*leve
   {
     use_polynomial_fit_ = config.use_polynomial_fit;
     NODELET_DEBUG ("[config_callback] Setting the use_polynomial_fit flag to: %d.", use_polynomial_fit_);
+#if PCL_VERSION_COMPARE(<, 1, 9, 0)
     impl_.setPolynomialFit (use_polynomial_fit_);
+#else
+    if (use_polynomial_fit_)
+    {
+      NODELET_WARN ("[config_callback] use_polynomial_fit is deprecated, use polynomial_order instead!");
+      if (impl_.getPolynomialOrder () < 2)
```
SteveMacenski, line 221: Rather than checking `< 2` why not use the `config.polynomial_order` already around?

I don't think the `else` needs to be there. Even the normal polynomial order API if the user asked for it didn't do that: https://github.com/ros-perception/perception_pcl/blob/acb450cc960694afc1e3d9cac525fc5063e71982/pcl_ros/src/pcl_ros/surface/moving_least_squares.cpp#L236

mvieth: I used [the code of setPolynomialFit](https://github.com/PointCloudLibrary/pcl/blob/master/surface/include/pcl/surface/mls.h#L360), so that the behaviour stays exactly the same.

SteveMacenski (approved):

pcl_ros/src/pcl_ros/surface/moving_least_squares.cpp
SteveMacenski: Thanks for the clarification

SteveMacenski:
Thanks for working with me on this. This same change should probably happen in Foxy-devel since its analog to noetic in OS / PCL versions

mvieth:
My pleasure! I created a second PR #305

PR #255, opened 2019-12-05
ROS 2 dashing pcl_ros port

sloretz:
This is still a work in progress (with a lot more work to do), but the Crop Box and Pass Through filters do work. It has everything from #229 plus changes from @mabelzhang's branch. I'm opening it for visibility.

It should build fine in Eloquent, but requires this rclcpp release to build in Dashing: https://github.com/ros/rosdistro/pull/23240

PR #306, opened 2020-08-31
Port the nodelet base class to a ROS2 component node

theseankelly:
Most of work comes from PR #255/#229 by sloretz/klintan


pcl_ros/include/pcl_ros/pcl_node.hpp
```diff
@@ -208,53 +246,39 @@ class PCLNodelet : public nodelet_topic_tools::NodeletLazy
...
+  /* \brief Return QoS settings for point cloud topic */
+  rclcpp::QoS
+  cloudQoS() const
+  {
+    rclcpp::QoS qos(max_queue_size_);
+    return qos;
   }
 
 public:
   EIGEN_MAKE_ALIGNED_OPERATOR_NEW
```
theseankelly, line 280: It's not clear to me if this is still needed.

theseankelly:
This doesn't add any actual functionality, there's no unit test, and the class is subject to change as other modules come online.

I've decided to split it out into its own PR so that work can proceed in parallel on the various sub-packages which derive from the base class

pcl_ros/include/pcl_ros/pcl_node.hpp
```diff
@@ -41,65 +41,103 @@
...
+  typedef ModelCoefficients::ConstSharedPtr ModelCoefficientsConstPtr;
+
+  /** \brief Constructor. */
+  PCLNode(std::string node_name, const rclcpp::NodeOptions & options)
+  : rclcpp::Node(node_name, options),
+    tf_buffer_(this->get_clock()),
+    tf_listener_(tf_buffer_)
+  {
+    {
+      rcl_interfaces::msg::ParameterDescriptor desc;
```
SteveMacenski, line 94: This is a really unusual way to declare parameters, can you explain why this is necessary? Why read only?

pcl_ros/include/pcl_ros/pcl_node.hpp
```diff
@@ -41,65 +41,103 @@
...
+  typedef pcl_msgs::msg::PointIndices PointIndices;
+  typedef PointIndices::SharedPtr PointIndicesPtr;
+  typedef PointIndices::ConstSharedPtr PointIndicesConstPtr;
+
+  typedef pcl_msgs::msg::ModelCoefficients ModelCoefficients;
+  typedef ModelCoefficients::SharedPtr ModelCoefficientsPtr;
+  typedef ModelCoefficients::ConstSharedPtr ModelCoefficientsConstPtr;
+
+  /** \brief Constructor. */
+  PCLNode(std::string node_name, const rclcpp::NodeOptions & options)
```
SteveMacenski, line 88: I think to be a component node, that the constructor can only take in options

pcl_ros/include/pcl_ros/pcl_node.hpp
theseankelly: To be honest, the reason is because I just took the implementation from pr #255 as a straight copy and didn't think much about it. I'll have to read up on how to declare parameters in ROS2 to do this right.

pcl_ros/include/pcl_ros/pcl_node.hpp
theseankelly: same comment as above

theseankelly:
too half baked.

Siddarth is taking over the porting of `filters` which will include this interface anyhow, so I'll let him push it through as appropriate.

siddhya:
I was trying this out along with enabling filter.cpp. The compilation is going through with a few changes on foxy release. If I just want to test if a some basic filter works what would be the easiest test case? Will I need to write it from scratch? No tests seem to be enabled as of now.

PR #311, opened 2020-10-18
Drop cmake_modules (not needed to find Eigen any longer)

jspricke:


SteveMacenski:
retriggering CI to see if build failures disappear from transient issues

SteveMacenski (approved):

PR #313, opened 2020-10-21
Change conversions of Vertices for new PCL versions

mvieth:
In https://github.com/PointCloudLibrary/pcl/commit/ad00c7bee2fad0391649479d90eee4461a2e74e7, the vertices field of pcl::Vertices changed from `std::vector<std::uint32_t>` to `std::vector<pcl::index_t>`, where index_t is an index type with configurable size (currently by default int). This commit makes conversions from and to pcl_msgs::Vertices possible again, moving the vector contents if possible.

### DIFF
--- a/pcl_conversions/include/pcl_conversions/pcl_conversions.h
+++ b/pcl_conversions/include/pcl_conversions/pcl_conversions.h
@@ -350,10 +350,25 @@ namespace pcl_conversions {
 
   /** pcl::Vertices <=> pcl_msgs::Vertices **/
 
+  namespace internal
+  {
+    template <class T>
+    inline void move(std::vector<T> &a, std::vector<T> &b)
+    {
+      b.swap(a);
+    }
+
+    template <class T1, class T2>
+    inline void move(std::vector<T1> &a, std::vector<T2> &b)
+    {
+      b.assign(a.cbegin(), a.cend());
+    }
+  }
+
   inline
   void fromPCL(const pcl::Vertices &pcl_vert, pcl_msgs::Vertices &vert)
   {
-    vert.vertices = pcl_vert.vertices;
+    vert.vertices.assign(pcl_vert.vertices.cbegin(), pcl_vert.vertices.cend());
   }
 
   inline
@@ -370,7 +385,7 @@ namespace pcl_conversions {
   inline
   void moveFromPCL(pcl::Vertices &pcl_vert, pcl_msgs::Vertices &vert)
   {
-    vert.vertices.swap(pcl_vert.vertices);
+    internal::move(pcl_vert.vertices, vert.vertices);
   }
 
   inline
@@ -387,7 +402,7 @@ namespace pcl_conversions {
   inline
   void toPCL(const pcl_msgs::Vertices &vert, pcl::Vertices &pcl_vert)
   {
-    pcl_vert.vertices = vert.vertices;
+    pcl_vert.vertices.assign(vert.vertices.cbegin(), vert.vertices.cend());
   }
 
   inline
@@ -404,7 +419,7 @@ namespace pcl_conversions {
   inline
   void moveToPCL(pcl_msgs::Vertices &vert, pcl::Vertices &pcl_vert)
   {
-    pcl_vert.vertices.swap(vert.vertices);
+    internal::move(vert.vertices, pcl_vert.vertices);
   }
 
   inline


pcl_conversions/include/pcl_conversions/pcl_conversions.h
```diff
@@ -350,10 +350,25 @@ namespace pcl_conversions {
...
+    template <class T>
+    inline void move(std::vector<T> &a, std::vector<T> &b)
+    {
+      b.swap(a);
+    }
+
+    template <class T1, class T2>
+    inline void move(std::vector<T1> &a, std::vector<T2> &b)
+    {
+      b.assign(a.cbegin(), a.cend());
```
kunaltyagi, line 364: IMO, this should be optional opt-in (via macro for example), or at least emit warnings when used. It doesn't `move` and for perf-reasons, this should not be done silently.

mvieth: Oh hi, I didn't know you visit this project regularly :smiley:
I did it like this so the `internal::move` with the `swap` would be used if possible, but if `pcl::index_t` is not `std::uint32_t`, the `swap` is not possible, so I do not see any other option than `assign`. I could add a `ROS_DEBUG` to inform the user of the disadvantage.

kunaltyagi: This issue will only crop up if a user has a custom PCL build (in which case, they'll be building perception_pcl too from source). That means we can force the user to opt-in (at compile or run-time). Using `assign` will cause a loss of information if the larger indices are truly needed.

Putting this behind `ROS_DEBUG` might not inform the user appropriately since debug messages are quite hidden, and require fiddling with rqt_logger_level or similar to find the root cause of error.

mvieth: Yes, right now it is only relevant for custom PCL master builds, but some day PCL 1.12 will be in the apt repos etc. and default for everyone. And then most people will not build perception_pcl from source.
I think an opt-in does not make much sense if there is no other way to use the function, and the slight performance disadvantage is IMO not important enough to explicitly require the user to say "yes".
Side question: how would you design an opt-in at run-time?
Ok yes, `ROS_DEBUG` is a bit hidden, but I am not sure how to generate a compile-time warning that triggers only if the specific function is used. `#warning` and `#pragma message` don't seem to be very portable and are always displayed, right? Maybe marking the move functions as deprecated would be good?
Regarding the other point (loss of information): yes, it would be safer to use a for-loop and check for each element if the receiving type is large enough for the value, but that would also be at the expense of performance.

kunaltyagi: I don't think copying entire vectors would be a slight performance disadvantage. quick-bench link: https://quick-bench.com/q/OYmVTj-rm2FLJtqFnrAHfsvj_pQ

> Regarding the other point (loss of information)

If the data comes from and exits using `pcl_msgs/Vertices.msg` only (and since the message definition is `uint32[] vertices`), there might be no loss of information. I haven't used this extensively, so I don't know if this is indeed the case.

> PCL 1.12 will be in the apt repos

The plan is to have `uint32_t` as the default type across all platforms. That doesn't break anything for ROS distributions via apt (or similar). If this changeset is for supporting custom builds of PCL, then it can be assumed that perception_pcl needs to be custom built too (the API+ABI will not be the same otherwise)

A quick grep shows no use if `fromPCL` for vertices, but `moveFromPCL` is used more than once at least. 

With this in mind, we can have a compile time flag: `PCL_ROS_FORCE_MOVE_INTO_COPY` or similar. The opt-in can't really be runtime, unless you're using env flags, in which case cross-platform compatibility would be bothersome. Same with pragma or warn messages. Deprecated would only cause confusion. I hope these rantings make some sense

In fact, this makes me wonder how perception_pcl is supported on windows (if it is ? 😕) since windows seems to have their int as `int64_t`

mvieth: > I don't think copying entire vectors would be a slight performance disadvantage. quick-bench link: https://quick-bench.com/q/OYmVTj-rm2FLJtqFnrAHfsvj_pQ

Ok, if you compare the relative performance of move and copy, then move is definitely much better, but I think using copy/assign instead of move won't create any performance difficulties for the user since there are probably functions in the user program that take much more time. I guess both versions are so fast (and only used very few times I assume) that it does not matter if copy takes 46 times longer than move. Anyways: the code _will_ use move if possible; but if move is not possible and copy is the only way, then there is no way but to accept that performance disadvantage.

> If the data comes from and exits using pcl_msgs/Vertices.msg only (and since the message definition is uint32[] vertices), there might be no loss of information. I haven't used this extensively, so I don't know if this is indeed the case.

That really depends on the use-case. The vertices could also be created by some pcl function and then converted to the ROS msg.

> The plan is to have uint32_t as the default type across all platforms. That doesn't break anything for ROS distributions via apt (or similar). If this changeset is for supporting custom builds of PCL, then it can be assumed that perception_pcl needs to be custom built too (the API+ABI will not be the same otherwise)

Wait, `pcl::index_t` will be `uint32_t`? The documentation say it will be `std::int32_t` (still signed) starting with 1.12: https://github.com/PointCloudLibrary/pcl/blob/c7e35fa26480e28d66867f39fde556410f64b0a1/common/include/pcl/types.h#L108
I really don't see how these changes would be for custom PCL builds only. I am sure they are important for _all_ builds, also for apt.

> A quick grep shows no use if fromPCL for vertices, but moveFromPCL is used more than once at least.

What code did you `grep`? The whole perception_pcl project? Can you link an example where `moveFromPCL` or `moveToPCL` is used for vertices? I didn't find one ...

> With this in mind, we can have a compile time flag: PCL_ROS_FORCE_MOVE_INTO_COPY or similar. The opt-in can't really be runtime, unless you're using env flags, in which case cross-platform compatibility would be bothersome. Same with pragma or warn messages. Deprecated would only cause confusion. I hope these rantings make some sense

The module pcl_conversions is header-only, so no compilation. That means that every project that uses the specific moveFrom/ToPCL function would have to set the flag.
IMO marking the function as deprecated would not be confusing. The message could be: `[[deprecated("Be advised that the vertices cannot be moved, they will be copied instead, which will give the same result, but is slower.")]] `. I think a warning is more appropriate here than an explicit opt-in.

> In fact, this makes me wonder how perception_pcl is supported on windows (if it is ? confused) since windows seems to have their int as int64_t

I only use Ubuntu, but [the ROS wiki](http://wiki.ros.org/melodic/Installation) lists windows as stable so perception_pcl should be supported? But why would int=int64_t be a problem for perception_pcl?

kunaltyagi: > But why would int=int64_t be a problem for perception_pcl?

Because the message definition for vertices is `unit32[] vertices`, and a copy is needed to happen or else it wouldn't work (hence your PR else your PR would not be needed)

> That means that every project that uses the specific moveFrom/ToPCL function would have to set the flag.

I did not think of this. That'd be painful. Also, a closer look at grep shows usage to be restricted only to the same header. No use of vertices or polygon mesh in pcl_ros.

The lack of use + the situation with windows tells me that the vertex related features were unknowingly broken before. No point in discussion about perf of a broken feature.

> Wait, pcl::index_t will be uint32_t? The documentation say it will be std::int32_t (still signed) starting with 1.12

My bad. It'd be `int32_t` as the documentation says.

> Anyways: the code will use move if possible; but if move is not possible and copy is the only way, then there is no way but to accept that performance disadvantage

I concur with this. I just wanted this cost to be made explicit to the user (more like the person compiling) somehow.

kunaltyagi: Please mark this as resolved.

SteveMacenski:
@kunaltyagi do you approve this?

@mvieth this fails noetic CI

mvieth:
> @mvieth this fails noetic CI

Only because of the hz-test which is generally unstable and would probably pass if re-run, and because of a compiler warning which was present long before this PR.

SteveMacenski:
toggle to retrigger

mvieth:
@SteveMacenski friendly ping

SteveMacenski:
@mvieth should this be in ROS2 as well given that's ROS moving forward (ROS1 deprecates in T-3 years) & will be what uses the modern PCL features.

mikepurvis:
I know there's only been one other change since the last ROS 1 release of this repo, but it would be great to get this released since the current release doesn't build against the current PCL (1.12.0, released July 2021).

SteveMacenski:
Do you just want Melodic? I added it to my sync list. I'll plan on doing something next week.

mikepurvis:
Noetic-only would be fine— it actually doesn't matter as we build our own binaries from source so we just need the tag to exist.

mikepurvis:
(But this will be an issue for anyone doing build-from-source stuff, like Homebrew for example bumped to PCL 1.12 a few weeks after it was released: https://github.com/Homebrew/homebrew-core/commits/master/Formula/pcl.rb)

PR #316, opened 2020-11-07
Change conversions of Vertices for new PCL versions

mvieth:
In https://github.com/PointCloudLibrary/pcl/commit/ad00c7bee2fad0391649479d90eee4461a2e74e7, the vertices field of pcl::Vertices changed from std::vector<std::uint32_t> to std::vector<pcl::index_t>, where index_t is an index type with configurable size (currently by default int). This commit makes conversions from and to pcl_msgs::Vertices possible again, moving the vector contents if possible.

#313 applied the same changes for melodic-devel

### DIFF
--- a/pcl_conversions/include/pcl_conversions/pcl_conversions.h
+++ b/pcl_conversions/include/pcl_conversions/pcl_conversions.h
@@ -356,10 +356,25 @@ namespace pcl_conversions {
 
   /** pcl::Vertices <=> pcl_msgs::Vertices **/
 
+  namespace internal
+  {
+    template <class T>
+    inline void move(std::vector<T> &a, std::vector<T> &b)
+    {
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

mikepurvis:
Thanks for this! It also patches cleanly onto the Melodic branch and is what is needed (along with 461a66ca0877b99e5f0cc0d6e157753db33dc0cd) to make ROS 1 pcl_ros work with PCL 1.12.

mvieth:
> Thanks for this! It also patches cleanly onto the Melodic branch and is what is needed (along with 461a66ca0877b99e5f0cc0d6e157753db33dc0cd) to make ROS 1 pcl_ros work with PCL 1.12.

No problem, you're welcome!

PR #330, opened 2021-06-17
Use complete namespace for traits members

mvieth:
This prevents errors like this: error: ‘name’ is not a member of ‘pcl::detail::traits’; did you mean ‘pcl::traits::name’?
The error appears for newer PCL versions (everything after commit https://github.com/PointCloudLibrary/pcl/commit/d39d3d3300746b952997e5bd2742dac7482aa5ab), but this change should also be fully compatible with older PCL versions.

### DIFF
--- a/pcl_ros/include/pcl_ros/point_cloud.h
+++ b/pcl_ros/include/pcl_ros/point_cloud.h
@@ -23,19 +23,19 @@ namespace pcl
 
       template<typename U> void operator() ()
       {
-        const char* name = traits::name<PointT, U>::value;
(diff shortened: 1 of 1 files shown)


mvieth:
Failing checks are unrelated (gcc warning and failed unit test)

SteveMacenski (approved):
LGTM

mvieth:
Shall I create a similar PR for ROS2? If yes, which branch exactly?

SteveMacenski:
`ros2` branch would be good enough for right now. You could also submit for Galactic and Foxy-devel if you liked, but not required

PR #331, opened 2021-06-24
shared library bug fix

pmusau17:
Signed-off-by: Patrick Musau <pmusau13ster@gmail.com>

Encountered the following error:
```
/usr/bin/ld: /opt/ros/galactic/lib/libpcl_ros_tf.a(transforms.cpp.o): relocation R_X86_64_PC32 against symbol `g_rcutils_logging_initialized' can not be used when making a shared object; recompile with -fPIC
```

 when trying to address this [PR](https://github.com/ros-perception/laser_filters/pull/104/).  This PR fixes that issue.

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -44,7 +44,7 @@ add_library(pcl_ros_tf src/transforms.cpp)
 ament_target_dependencies(pcl_ros_tf
   ${dependencies}
 )
-target_link_libraries(pcl_ros_tf ${PCL_LIBRARIES})
+target_link_libraries(pcl_ros_tf SHARED ${PCL_LIBRARIES})
 
 ### Nodelets
 #


pmusau17:
@SteveMacenski friendly ping for review.

SteveMacenski (approved):

SteveMacenski:
Thanks!

jonbinney:
I think this doesn't work properly. laser_filters (which links to pcl_ros) fails to build after this commit with the cryptic error: `/usr/bin/ld: cannot find -lSHARED`. Reverting this commit fixes the build. I'm not sure exactly how the list of libraries is being passed around to cause this though.
(this is on Focal Fossa with cmake 3.16.3 and ROS Foxy)

You should be able to reproduce the problem by compiling perception_pcl and laser_filters (at git commit b2845c96cc4fef2cf381c12676ab290cda80d085 ) in the same workspace using colcon on foxy. Ironically I ran into this problem right when I was intending to remove pcl_ros as a dependency from laser_filters, since pcl pulls in 500MB of dependencies (looks like it needs a bunch of graphical stuff).

SteveMacenski:
Does that happen when you build in a clean workspace or clear out the caches in install/build spaces? I'm wondering if you're seeing that simply from information hanging around from prior builds. 

I build locally in a brand new workspace with foxy binaries (pulling in the current `ros2` branches of these 2 repos) and I don't get that error. I get a 

```
Finished <<< perception_pcl [0.80s]                                                            
--- stderr: laser_filters                              
In file included from /tmp/ws/src/laser_filters/src/laser_scan_filters.cpp:30:
/tmp/ws/src/laser_filters/include/laser_filters/median_filter.h:42:10: fatal error: filters/median.hpp: No such file or directory
   42 | #include "filters/median.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~

```

but I assume I wouldn't even get to that point if that was an issue.

But I'm happy to revert if we find that yes it is indeed an issue.

jonbinney:
I get this error even when starting from a clean workspace.

You're seeing a compiler error, so I don't think you're getting to the linker step where I'm seeing an error. The error you're getting is a bug that has been fixed in the filters package (by you! https://github.com/ros/filters/pull/50 ) but that fix hasn't been released into foxy yet.

Could you try adding the filters package with the fix into your workspace and see if you can replicate the bug?

jonbinney:
I think the issue is that pcl_ros ends up including a library called SHARED in its INTERFACE_LINK_LIBRARIES cmake variable. Running `cat install/pcl_ros/share/pcl_ros/cmake/export_pcl_rosExport.cmake|grep SHARED` gives:
```
INTERFACE_LINK_LIBRARIES "rclcpp::rclcpp;sensor_msgs::sensor_msgs__rosidl_generator_c;sensor_msgs::
... (lots of stuff)
linux-gnu/libtiff.so;/usr/lib/x86_64-linux-gnu/libexpat.so;/usr/lib/x86_64-linux-gnu/libpcl_common.so
;SHARED;pcl_common;pcl_kdtree;pcl_octree;pcl_search;pcl_features;pcl_sample_consensus;pcl_filters;pcl
```
In cmake's documentation for target_link_libraries, I don't see any valid use for "SHARED".

SteveMacenski:
Fair enough, yes I do see it now (sorry about that), however removing it actually gives me the error the user initial reported that this PR fixed. 

I fixed that issue though doing what the error suggested: "recompile with -fPIC". I'm not sure why @pmusau17 didn't choose to go down that route. Lets revert this PR and I'll open up another one adding the `-fPIC` compiler flags. That should do the trick for everyone.

SteveMacenski:
@jonbinney can you verify this fixes your concern https://github.com/ros-perception/perception_pcl/pull/335

jonbinney:
Yep, that does the trick! Thanks @SteveMacenski !

SteveMacenski:
No worries, thanks for pointing it out

PR #346, opened 2021-09-22
pcl_ros: Use vec.data() instead of &vec[0]

### LINKED ISSUE #333: Undefined behavior in point_cloud.h
The point cloud serialization code in `point_cloud.h` relies on undefined behavior when using `operator[]()` to access element 0 of a vector when the vector is empty, as implemented here (and potentially elsewhere):
https://github.com/ros-perception/perception_pcl/blob/bd21be2cb8ec6ad70a282c61b14aa9dd43edf9ff/pcl_ros/include/pcl_ros/point_cloud.h#L167

C++11 [added a data()](http://www.open-std.org/jtc1/sc22/wg21/docs/lwg-defects.html#464) method to `std::vector` to avoid this issue - `point_cloud.h` should be updated to use `.data()` instead of a reference to the first vector element to get data pointers from std::vectors.  Alternately, a size check should be performed before accessing `vector[0]`.

richmattes:
Update pcl_ros to use the data() member function of STL containers to
get the pointer to the underlying storage, instead of dereferencing the
zeroth element and taking its reference.  When a container is of size 0,
dereferencing element 0 with operator[]() is undefined behavior, and
will trigger assertions when features like _GLIBCXX_ASSERTIONS are
enabled.

Fixes #333.

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -1,8 +1,14 @@
 cmake_minimum_required(VERSION 3.0.2)
 project(pcl_ros)
 
-set(CMAKE_VERBOSE_MAKEFILE ON)
-add_compile_options(-std=c++14)
+# CMake 3.1 added support for CMAKE_CXX_STANDARD to manage the C++ standard version
+# Use CMake C++ standard management where possible for better interoperability
+# with dependencies that export target standards and/or features
+if (${CMAKE_VERSION} VERSION_LESS "3.1")
+  add_compile_options(-std=c++14)
+else()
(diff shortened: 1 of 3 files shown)


richmattes:
I went ahead and fixed the compiler warning issue on Buster, CI all passes now.  I'd appreciate it if this got merged and released for melodic and noetic soon.

andrewcapodieci:
I'm experiencing a related issue with publishing of an empty point cloud. This change fixes the problem.

SteveMacenski (approved):
Please submit for `ros2` as well and I'll merge'm both. Been on PTO, catching up now.

PR #354, opened 2022-01-26
Install headers to include/${PROJECT_NAME}

sloretz:
Part of ros2/ros2#1150 - this installs headers to another folder to avoid include directory search order issues when overriding these packages.


jacobperron (approved):

SteveMacenski (approved):
Whoops, sorry this got mixed up in my queue

PR #359, opened 2022-04-01
Migrate pcd to pointcloud tool

ivanpauno:
To test:

- Download https://github.com/PointCloudLibrary/data/blob/master/biwi_face_database/model.pcd.
- Run `ros2 run pcl_ros pcd_to_pointcloud --ros-args -p file_name:=$(cd . && pwd)/model.pcd`
- To check the point cloud is being sent periodically `ros2 topic echo /cloud_pcd` (or use any tool to visualize the pointcloud).

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -11,7 +11,6 @@ endif()
 ## Find system dependencies
 find_package(Eigen3 REQUIRED)
 find_package(PCL REQUIRED QUIET COMPONENTS core common features filters io segmentation surface)
-message(WARNING "components: ${PCL_COMPONENTS}")
 
 ## Find ROS package dependencies
 find_package(ament_cmake REQUIRED)
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -130,7 +130,7 @@ ament_target_dependencies(pcd_to_pointcloud_lib
   sensor_msgs
   pcl_conversions)
 rclcpp_components_register_node(pcd_to_pointcloud_lib
-  PLUGIN "pcl_ros::PCDGenerator"
+  PLUGIN "pcl_ros::PCDPublisher"
   EXECUTABLE pcd_to_pointcloud)
 #
 #add_executable(pointcloud_to_pcd tools/pointcloud_to_pcd.cpp)
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -120,7 +120,6 @@ ament_target_dependencies(pcl_ros_tf
 ### Tools
 #
 add_library(pcd_to_pointcloud_lib SHARED tools/pcd_to_pointcloud.cpp)
-# add_executable(pcd_to_pointcloud tools/pcd_to_pointcloud.cpp)
 target_link_libraries(pcd_to_pointcloud_lib
   ${PCL_LIBRARIES})
 target_include_directories(pcd_to_pointcloud_lib PUBLIC
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -10,12 +10,14 @@ endif()
 
 ## Find system dependencies
 find_package(Eigen3 REQUIRED)
-find_package(PCL REQUIRED QUIET COMPONENTS core features filters io segmentation surface)
+find_package(PCL REQUIRED QUIET COMPONENTS core common features filters io segmentation surface)
+message(WARNING "components: ${PCL_COMPONENTS}")
 
 ## Find ROS package dependencies
 find_package(ament_cmake REQUIRED)
 find_package(pcl_conversions REQUIRED)
 find_package(rclcpp REQUIRED)
+find_package(rclcpp_components REQUIRED)
 find_package(sensor_msgs REQUIRED)
 find_package(geometry_msgs REQUIRED)
 find_package(tf2 REQUIRED)
@@ -118,8 +120,20 @@ ament_target_dependencies(pcl_ros_tf
 #
 ### Tools
 #
-#add_executable(pcd_to_pointcloud tools/pcd_to_pointcloud.cpp)
-#target_link_libraries(pcd_to_pointcloud ${Boost_LIBRARIES} ${catkin_LIBRARIES} ${PCL_LIBRARIES})
+add_library(pcd_to_pointcloud_lib SHARED tools/pcd_to_pointcloud.cpp)
+# add_executable(pcd_to_pointcloud tools/pcd_to_pointcloud.cpp)
+target_link_libraries(pcd_to_pointcloud_lib
+  ${PCL_LIBRARIES})
+target_include_directories(pcd_to_pointcloud_lib PUBLIC
+  ${PCL_INCLUDE_DIRS})
+ament_target_dependencies(pcd_to_pointcloud_lib
+  rclcpp
+  rclcpp_components
+  sensor_msgs
+  pcl_conversions)
+rclcpp_components_register_node(pcd_to_pointcloud_lib
+  PLUGIN "pcl_ros::PCDGenerator"
+  EXECUTABLE pcd_to_pointcloud)
 #
 #add_executable(pointcloud_to_pcd tools/pointcloud_to_pcd.cpp)
 #target_link_libraries(pointcloud_to_pcd ${Boost_LIBRARIES} ${catkin_LIBRARIES} ${PCL_LIBRARIES})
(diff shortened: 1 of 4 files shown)


pcl_ros/CMakeLists.txt
```diff
@@ -118,8 +119,20 @@ ament_target_dependencies(pcl_ros_tf
 #
 ### Tools
 #
-#add_executable(pcd_to_pointcloud tools/pcd_to_pointcloud.cpp)
-#target_link_libraries(pcd_to_pointcloud ${Boost_LIBRARIES} ${catkin_LIBRARIES} ${PCL_LIBRARIES})
+add_library(pcd_to_pointcloud_lib SHARED tools/pcd_to_pointcloud.cpp)
+# add_executable(pcd_to_pointcloud tools/pcd_to_pointcloud.cpp)
```
SteveMacenski: Does this need to be here? I don't see a `main` anymore for that to be possible. 

Though it might be nice to add a new main in a new file so that there is a standalone version available

SteveMacenski: Does this need to be here? I don't see a `main` anymore for that to be possible. 

Though it might be nice to add a new main in a new file so that there is a standalone version available

pcl_ros/tools/pcd_to_pointcloud.cpp
```diff
@@ -43,126 +43,89 @@
...
     // Maximum number of outgoing messages to be queued for delivery to subscribers = 1
 
     cloud_topic_ = "cloud_pcd";
-    pub_.advertise(nh_, cloud_topic_.c_str(), 1);
-    private_nh_.param("frame_id", tf_frame_, std::string("/base_link"));
-    ROS_INFO(
-      "Publishing data on topic %s with frame_id %s.", nh_.resolveName(
-        cloud_topic_).c_str(), tf_frame_.c_str());
-  }
+    tf_frame_ = this->declare_parameter("spin_period", tf_frame_);
```
SteveMacenski: spin period is tf frame?

SteveMacenski: spin period is tf frame?

pcl_ros/tools/pcd_to_pointcloud.cpp
```diff
@@ -43,126 +43,89 @@
...
+      throw std::runtime_error{"could not open pcd file"};
     }
     cloud_.header.frame_id = tf_frame_;
-    return 0;
+    int nr_points = cloud_.width * cloud_.height;
+
+    auto fields_list = pcl::getFieldsList(cloud_);
+    auto resolved_cloud_topic =
+      this->get_node_topics_interface()->resolve_topic_name(cloud_topic_);
+    RCLCPP_DEBUG(
```
SteveMacenski: Why not make this INFO and then delete the INFO below that is a partial subset of what this says?

SteveMacenski: Why not make this INFO and then delete the INFO below that is a partial subset of what this says?

pcl_ros/tools/pcd_to_pointcloud.cpp
ivanpauno: lol, will fix that

ivanpauno:
Thanks for the review @SteveMacenski !
Addressed your comments in https://github.com/ros-perception/perception_pcl/pull/359/commits/31a7fc0570bccc531a8b310f9375ae7ad4be9c0b.

pcl_ros/CMakeLists.txt
```diff
@@ -118,8 +119,19 @@ ament_target_dependencies(pcl_ros_tf
...
+  ${PCL_LIBRARIES})
+target_include_directories(pcd_to_pointcloud_lib PUBLIC
+  ${PCL_INCLUDE_DIRS})
+ament_target_dependencies(pcd_to_pointcloud_lib
+  rclcpp
+  rclcpp_components
+  sensor_msgs
+  pcl_conversions)
+rclcpp_components_register_node(pcd_to_pointcloud_lib
+  PLUGIN "pcl_ros::PCDGenerator"
```
jacobperron: nitpick: I would make the name of the plugin more closely resemble the name of the executable. I don't think `PCDGenerator` is an accurate description for what it does.

ivanpauno: I picked the old name used in ROS 1.
If we want to update the name, `PCDPublisher` may be a good one.

@SteveMacenski what do you think?

SteveMacenski: Yeah, more in line with the file name would be good

pcl_ros/CMakeLists.txt
```diff
@@ -118,8 +119,19 @@ ament_target_dependencies(pcl_ros_tf
 #
 ### Tools
 #
-#add_executable(pcd_to_pointcloud tools/pcd_to_pointcloud.cpp)
-#target_link_libraries(pcd_to_pointcloud ${Boost_LIBRARIES} ${catkin_LIBRARIES} ${PCL_LIBRARIES})
+add_library(pcd_to_pointcloud_lib SHARED tools/pcd_to_pointcloud.cpp)
+target_link_libraries(pcd_to_pointcloud_lib
+  ${PCL_LIBRARIES})
+target_include_directories(pcd_to_pointcloud_lib PUBLIC
```
SteveMacenski, line 125: I meant that more as a suggestion to actually add a `main` file option so that users don't have to only interact with this as a component (e.g. publish me a pointcloud for a test). I try to have any components also have an executable counterpart so that users can use either that they wish that is most sensible for their workflow.

jacobperron: `rclcpp_components_register_node` also creates a standalone executable (by generating a main function). So you can invoke the program like: `ros2 run pcl_ros pcd_to_pointcloud`.

SteveMacenski: Oh ok, I actually did not know that. Is that documented somewhere / new?

jacobperron: It's documented [as part of the docstring](https://github.com/ros2/rclcpp/blob/03fa731d2345c41e5a4e436bf0730012ece96ce5/rclcpp_components/cmake/rclcpp_components_register_node.cmake#L15-L16). The feature was added in 2019, so not terribly new. I'm not sure if we call it out in any tutorials, but a lot of our demo packages use it.

SteveMacenski: Ah feels like something that should be explicitly mentioned in composition tutorials. Here nor there for this PR, but I'd suggest maybe spending a few minutes if you can think of a place that this note would be good to expose to users. Many people's complaints about ROS 2 are these "hidden" features or got-ya's that aren't documented. Small actions accumulated over time make a big difference :-) 

Anyhow, no action here on that needed. I think the renaming is the only thing left.

jacobperron: Unfortunately, I don't think we have a tutorial on writing and building node components. I'm only aware of [this tutorial](https://docs.ros.org/en/rolling/Tutorials/Composition.html) for interacting with components that already exist on the system.

SteveMacenski: ... there's no "how do you make a component" tutorial?

SteveMacenski: I suppose https://docs.ros.org/en/rolling/Concepts/About-Composition.html?highlight=rclcpp_components_register_node#writing-a-component and https://docs.ros.org/en/rolling/Tutorials/Actions/Writing-a-Cpp-Action-Server-Client.html?highlight=rclcpp_components_register_node#writing-an-action-server-and-client-c are as good of a place as currently exists.

pcl_ros/CMakeLists.txt
ivanpauno: I think I have already removed that line in https://github.com/ros-perception/perception_pcl/pull/359/commits/31a7fc0570bccc531a8b310f9375ae7ad4be9c0b.

pcl_ros/CMakeLists.txt
ivanpauno: Yeah, as @jacobperron mentions, this already is creating an executable.
The [PR description](https://github.com/ros-perception/perception_pcl/pull/359#issue-1190196659) has some instructions of how to use it.

> I suppose https://docs.ros.org/en/rolling/Concepts/About-Composition.html?highlight=rclcpp_components_register_node#writing-a-component and https://docs.ros.org/en/rolling/Tutorials/Actions/Writing-a-Cpp-Action-Server-Client.html?highlight=rclcpp_components_register_node#writing-an-action-server-and-client-c are as good of a place as currently exists.

Yeah, I agree this should be explicitly mentioned in some of the tutorials.
Currently we're using it in many of the demos though, https://github.com/ros2/demos/blob/f008dbbeefd7da30c4133ad67655554d9f7f8519/demo_nodes_cpp/CMakeLists.txt#L112-L114.

pcl_ros/CMakeLists.txt
ivanpauno: Renamed in https://github.com/ros-perception/perception_pcl/pull/359/commits/467ba51b03456b64f39dd8c50317965fafb8f2e2.

SteveMacenski (approved):
LGTM

SteveMacenski:
@jacobperron you good with it?

jacobperron (approved):
LGTM! thanks!

lukicdarkoo:
This is super useful, thanks! Can we get the package published?

PR #361, opened 2022-04-13
fix: use common header directly

wep21:
fix buid error as below on jammy
```
In file included from /home/daisuke/workspace/sandbox_ws/src/perception_pcl/pcl_ros/tools/pcd_to_pointcloud.cpp:52:
/usr/include/pcl-1.12/pcl/io/io.h:41:22: error: expected constructor, destructor, or type conversion before ‘(’ token
   41 | PCL_DEPRECATED_HEADER(1, 15, "Please include pcl/common/io.h directly.")
      |                      ^
In file included from /usr/include/pcl-1.12/pcl/io/io.h:42,
                 from /home/daisuke/workspace/sandbox_ws/src/perception_pcl/pcl_ros/tools/pcd_to_pointcloud.cpp:52:
/usr/include/pcl-1.12/pcl/common/io.h: In function ‘std::string pcl::getFieldsList(const pcl::PCLPointCloud2&)’:
/usr/include/pcl-1.12/pcl/common/io.h:110:17: error: ‘accumulate’ is not a member of ‘std’
  110 |     return std::accumulate(std::next (cloud.fields.begin ()), cloud.fields.end (), cloud.fields[0].name,
```

### DIFF
--- a/pcl_ros/src/pcl_ros/features/feature.cpp
+++ b/pcl_ros/src/pcl_ros/features/feature.cpp
@@ -47,7 +47,7 @@
 // #include "pfh.cpp"
 // #include "principal_curvatures.cpp"
 // #include "vfh.cpp"
-#include <pcl/io/io.h>
+#include <pcl/common/io.h>
 #include <message_filters/null_types.h>
 #include <vector>
 #include "pcl_ros/features/feature.hpp"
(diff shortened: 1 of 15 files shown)


wep21:
cc @SteveMacenski

mvieth:
:+1: from me, pcl/io/io.h contains only the redirect to pcl/common/io.h since 2011, and the deprecation notice since May 2021 (https://github.com/PointCloudLibrary/pcl/commits/master/io/include/pcl/io/io.h). However even including pcl/common/io.h might be unnecessary in some places, identifying and removing unnecessary includes with include-what-you-use could be an option.

SteveMacenski (approved):

PR #369, opened 2022-05-29
Rename headers for humble

smnogar:
Humble has standardized a lot of standard packages on using .hpp for headers. This will break on galactic, so recommend pushing this to a new humble branch (or just make sure others are aware ros2 targets humble).

### DIFF
--- a/pcl_ros/src/pcl_ros/features/boundary.cpp
+++ b/pcl_ros/src/pcl_ros/features/boundary.cpp
@@ -35,7 +35,7 @@
  *
  */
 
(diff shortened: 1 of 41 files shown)


SteveMacenski (approved):

SteveMacenski:
https://build.ros2.org/job/Rpr__perception_pcl__ubuntu_jammy_amd64/1/console

The linters require the headers are also reordered, since now there are `hpp` files before system headers. Control+F for `cpplint` and it shows failures in many of the files changed due to this PR

PR #380, opened 2022-11-18
Avoid redundant boost::make_shared copy in pcl_ros::Publisher<PointT>

ardiya:
`boost make_shared` is actually making a copy. This PR introduces the `sensor_msgs::PointCloud2` as pointer directly so that it doesn't need a `boost::make_shared` convertsion to pointer anymore.


SteveMacenski (approved):
Thanks!

PR #384, opened 2022-11-29
use modular pcl dependencies for bloom

Aposhian:
Now that rosdep has [rules for the modular debian and ubuntu PCL packages](https://github.com/ros/rosdistro/pull/35289), those should be used here.

This doesn't make a big difference for a rosdep + colcon workflow, but it makes a _big_ difference for the bloomed debian packages. This is because `libpcl-dev` pulls in all kinds of things, including GUI tools, java, you name it. So it should only be pulled in as a build dependency (to get the headers) and then at runtime we should just pull in the .so's we need with the modular packages.

Note that for platforms that don't have modular pcl packages, these rosdep rules point to the regular pcl packages anyway, so this shouldn't break anything for them.

### DIFF
--- a/pcl_conversions/package.xml
+++ b/pcl_conversions/package.xml
@@ -21,13 +21,17 @@
 
   <buildtool_depend>ament_cmake</buildtool_depend>
 
+  <build_depend>libpcl-all-dev</build_depend>
+
   <depend>eigen</depend>
-  <depend>libpcl-all-dev</depend>
   <depend>message_filters</depend>
   <depend>pcl_msgs</depend>
   <depend>rclcpp</depend>
   <depend>sensor_msgs</depend>
   <depend>std_msgs</depend>
+
+  <exec_depend>libpcl-common</exec_depend>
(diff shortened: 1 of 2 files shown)


SteveMacenski (approved):

agoeckner:
Hey @SteveMacenski, has this been released to Humble yet? It seems that ros-humble-perception-pcl brings in a lot of extra packages.

jdlangs:
This change doesn't seem to be reflected in any debians from the build farm. It would be nice to get this released as it leads to a noticeable improvement in production application image size.

agoeckner:
> This change doesn't seem to be reflected in any debians from the build farm. It would be nice to get this released as it leads to a noticeable improvement in production application image size.

I'm also not seeing this in the Humble binaries.

Yadunund:
Looks like not exporting `libpcl-all-dev` is leading to regressions in the ROS buildfarm https://github.com/ros-perception/perception_pcl/issues/446.

PR #385, opened 2022-11-30
migrate pcl_nodelet.hpp to pcl_node.hpp

daisukes:
@SteveMacenski 

This is the initial PR of a series of PRs to migrate filters.
The previous PR #383 was about 2400 (+1600, -800) changes.
You guessed it would take 4 hours to review and asked me to break them down so that you can review each of them in about 15 mins.
So, I assume that about 150 is an appropriate size and make this PR.

However, this is just a header file, so if you want to have a minimal working code, please let me know. I will commit the following changes too.
```
 pcl_ros/CMakeLists.txt                              |  22 +++--
 pcl_ros/include/pcl_ros/filters/extract_indices.hpp |  26 +++---
 pcl_ros/include/pcl_ros/filters/filter.hpp          |  69 +++++++--------
 pcl_ros/src/pcl_ros/filters/extract_indices.cpp     |  61 ++++++++-----
 pcl_ros/src/pcl_ros/filters/filter.cpp              | 190 +++++++++++++++++-----------------------
 5 files changed, 177 insertions(+), 191 deletions(-)
```

To minimize changes, I have not included 
- lazy subscription
- format fix by cpplint and uncrustify

PR #383, opened 2022-11-29
ROS2 port pcl_ros/filters

daisukes:
This is a ros2 porting of pcl_ros filters.

I have been working on my ROS2 migration project and found that most of the pcl_ros functions are not ported yet.
https://discourse.ros.org/t/call-for-help-pcl-ros-for-ros2/15131

I have just ported filters because I need them, and they could be ported without type adaptation [REP-2007](https://ros.org/reps/rep-2007.html).
```
sub_input_ = pnh_->subscribe<sensor_msgs::PointCloud2> (
  "input", max_queue_size_,  bind (&Filter::input_indices_callback, this, _1, pcl_msgs::PointIndicesConstPtr ()));
```
https://github.com/ros-perception/perception_pcl/blob/melodic-devel/pcl_ros/src/pcl_ros/filters/filter.cpp#L130

## contributions
This work is based on #255
- all filters (node/component)
- lazy subscription (based on [a workaround](https://github.com/mikeferguson/ros2_cookbook/blob/main/rclcpp/workarounds.md))
- integration tests for filters

PR #388, opened 2022-12-12
migrate abstract filter node

daisukes:
This is the second PR of the migration PR series (#383)
The previous PR is #385 

- migrate filter.{cpp,hpp}
- CMakeLists.txt related to those files

### DIFF
--- a/pcl_ros/include/pcl_ros/filters/filter.hpp
+++ b/pcl_ros/include/pcl_ros/filters/filter.hpp
@@ -63,11 +63,14 @@ public:
   /** \brief Filter constructor
     * \param node_name node name
     * \param options node options
-    * \param use_frame_params declare and handle parameter callback for input_frame and output_frame if true
     */
-  Filter(std::string node_name, const rclcpp::NodeOptions & options, bool use_frame_params=false);
+  Filter(std::string node_name, const rclcpp::NodeOptions & options);
 
 protected:
+  /** \brief declare and subscribe to param callback for input_frame and output_frame params */
+  void
+  use_frame_params();
+
   /** \brief The input PointCloud subscriber. */
   rclcpp::Subscription<PointCloud2>::SharedPtr sub_input_;
 
--- a/pcl_ros/include/pcl_ros/filters/filter.hpp
+++ b/pcl_ros/include/pcl_ros/filters/filter.hpp
@@ -39,10 +39,10 @@
 #define PCL_ROS__FILTERS__FILTER_HPP_
 
 #include <pcl/filters/filter.h>
-#include <dynamic_reconfigure/server.h>
+#include <memory>
 #include <string>
-#include "pcl_ros/pcl_nodelet.hpp"
-#include "pcl_ros/FilterConfig.hpp"
+#include <vector>
+#include "pcl_ros/pcl_node.hpp"
 
 namespace pcl_ros
 {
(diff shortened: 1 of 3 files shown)


pcl_ros/include/pcl_ros/filters/filter.hpp
```diff
@@ -52,19 +52,24 @@ namespace sync_policies = message_filters::sync_policies;
...
   typedef pcl::IndicesPtr IndicesPtr;
   typedef pcl::IndicesConstPtr IndicesConstPtr;
 
-  Filter() {}
+  /** \brief Filter constructor
+    * \param node_name node name
+    * \param options node options
+    * \param use_frame_params declare and handle parameter callback for input_frame and output_frame if true
+    */
+  Filter(std::string node_name, const rclcpp::NodeOptions & options, bool use_frame_params=false);
```
SteveMacenski: To be a composable node, the only constructor fields can be the NodeOptions -- is this a node to be deployed itself? If so, its probably wise to use that so that all of these can be composed into a single process to be able to leverage IPC optimizations

pcl_ros/src/pcl_ros/filters/filter.cpp
```diff
@@ -150,114 +153,121 @@ pcl_ros::Filter::unsubscribe()
...
-  PCLNodelet::onInit();
+  pub_output_ = create_publisher<PointCloud2>("output", max_queue_size_);
 
-  // Call the child's local init
-  bool has_service = false;
-  if (!child_init(*pnh_, has_service)) {
-    NODELET_ERROR("[%s::onInit] Initialization failed.", getName().c_str());
-    return;
-  }
+  if (use_frame_params) {
```
SteveMacenski: I'm not sure I understand the use frame params as a porting change. What's going on here that should go into this base class?

pcl_ros/include/pcl_ros/filters/filter.hpp
SteveMacenski: If not, this is only a base node, that's fine, but just an FYI that this could make some odd looking derived nodes if they have to call the constructor `Filter` in their constructors with hardcoded values of `use_frame_params` (since the others are passed `NodeOptions` or can be hardcoded `node_name`)

daisukes: > is this a node to be deployed itself?

No, this is an abstract class and not deployed.

> If not, this is only a base node, that's fine, but just an FYI that this could make some odd looking derived nodes if they have to call the constructor `Filter` in their constructors with hardcoded values of `use_frame_params` (since the others are passed `NodeOptions` or can be hardcoded `node_name`)

The class has a callback function that handles `dynamic_reconfigure` parameter change for `input_frame` and `output_frame`, however all sub classes do not use this callback and just copy the logic ([comment](https://github.com/ros-perception/perception_pcl/blob/melodic-devel/pcl_ros/src/pcl_ros/filters/voxel_grid.cpp#L119) in the original code).

`use_frame_params` is a replacement of `has_service` flag in `child_init` function, which is not actually worked well in the original code, because all sub class set it to true or override `child_init` itself.

pcl_ros/src/pcl_ros/filters/filter.cpp
daisukes: 3 out of 6 filters use `input_frame` and `output_frame` as their parameters.
As the original author tried to eliminate copied logic as mentioned the above comment, I put this here.

pcl_ros/include/pcl_ros/filters/filter.hpp
SteveMacenski: But should `use_frame_params` be a constructor argument?

pcl_ros/src/pcl_ros/filters/filter.cpp
SteveMacenski: See comment above, I don't have issue with it being a param, but a constructor argument feels odd

daisukes:
@SteveMacenski Thank you for the feedback. I move it to `use_frame_params` function.

SteveMacenski (approved):

PR #393, opened 2023-01-18
Adding separate leaf sizes for each axis in the VoxelGrid filter

ayrton04:
I needed to have a different leaf size for each axis (particularly `z`), but the `VoxelGrid` filter currently assumes uniform leaf sizes. The behavior here defaults to the current behavior, but optionally allows for each axis leaf size to be overridden.


ayrton04:
The Noetic PR builds succeeded. Failure in the Melodic PR build seems to be an unrelated test.

paulbovbel (approved):

PR #398, opened 2023-02-10
Migrate the ROS1 pcl::VoxelGrid filter to ROS2 pcl_ros::VoxelGrid

asymingt:
## Overview

This PR migrates the VoxelGrid filter. The default parameter ranges and values were inherited from the hard-coded values in the ROS1 `.cfg` file for dynamic_reconfigure. Two things to note:

1. Note that I have changed the namespace from `pcl::` to `pcl_ros::` to be consistent with the other filters in the package. I don't expect this will be a breaking change because ROS1 packages will need to be hand-migrated to ROS2, and there's good benefit to having a consistent namespacing across the package.
2. I have removed the `keep_organized` parameter from common parameters in `filter.cpp` to the `Passthrough` where it is used, thereby allowing and the `VoxelGrid` and `Passthrough` to both use the common parameters. 

## Build

CI linter is surfacing `cpplint` and `uncrustify` errors from different files than those touched by this PR.

### DIFF
--- a/pcl_ros/include/pcl_ros/filters/voxel_grid.hpp
+++ b/pcl_ros/include/pcl_ros/filters/voxel_grid.hpp
@@ -40,10 +40,8 @@
 
 // PCL includes
 #include <pcl/filters/voxel_grid.h>
+#include <vector>
 #include "pcl_ros/filters/filter.hpp"
-
-// Dynamic reconfigure
-#include "pcl_ros/VoxelGridConfig.hpp"
 
 namespace pcl_ros
 {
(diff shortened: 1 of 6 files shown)


pcl_ros/include/pcl_ros/pcl_node.hpp
```diff
@@ -201,7 +201,7 @@ class PCLNode : public rclcpp::Node
   /** \brief TF listener object. */
   tf2_ros::TransformListener tf_listener_;
   tf2_ros::Buffer tf_buffer_;
-
```
asymingt, line 204: Courtesy of `ament_uncrustify --reformat` :man_shrugging:

SteveMacenski:
@daisukes I don't suppose you can do me a solid and co-review these PRs with me? 4 eyes are better than 2 :smile:

pcl_ros/src/pcl_ros/filters/filter.cpp
```diff
@@ -247,20 +247,11 @@ pcl_ros::Filter::add_common_params()
     "Set to true if we want to return the data outside [filter_limit_min; filter_limit_max].";
   declare_parameter(flneg_desc.name, rclcpp::ParameterValue(false), flneg_desc);
 
-  rcl_interfaces::msg::ParameterDescriptor keep_organized_desc;
```
asymingt, line 250: We don't need this in `VoxelGrid` so let's move it to the constructor for `Passthrough`.

daisukes: I agree with this

daisukes (commented):
LGTM

- This includes a recent change by PR #393 too.

asymingt:
> * #393

Correct -- I tried to parity match feature-wise with the tip of the ROS1 branch to make transition to ROS2 easier for folk.

pcl_ros/include/pcl_ros/filters/voxel_grid.hpp
```diff
@@ -53,38 +51,42 @@ namespace pcl_ros
...
     * \param output the resultant filtered dataset
     */
-  virtual void
+  inline void
   filter(
-    const PointCloud2::ConstPtr & input, const IndicesPtr & indices,
-    PointCloud2 & output);
+    const PointCloud2::ConstSharedPtr & input, const IndicesPtr & indices,
+    PointCloud2 & output) override
+  {
```
SteveMacenski, line 63: move to cpp file

pcl_ros/src/pcl_ros/filters/voxel_grid.cpp
```diff
@@ -35,100 +35,213 @@
...
+          "Setting the minimum points per voxel to: %u.",
+          minPointsPerVoxel);
+        impl_.setMinimumPointsNumberPerVoxel(param.as_int());
+      }
+    }
+    // Defer decision on final leaf size until after all parameters are collected.
+    if (param.get_name() == "leaf_size") {
+      leaf_size.setConstant(param.as_double());
+    }
+    if (param.get_name() == "leaf_size_x") {
```
SteveMacenski: This leaf size thing is very convoluted. You have this leaf_size array sized 3 you copy above. Why not just populate that here? You then check if `impl_.getLeafSize() != leaf_size` below, which would still trigger the same way without these additional 2nd level indirections

daisukes: I think we cannot assume the order of parameter iteration. If we have both `leaf_size` and `leaf_size_x`, the result can be nondeterministic.
How about using the original structure? Just putting the parameter values into `config_leaf_size`, `config_leaf_size_x`, ... in the loop and reuse the same logic here after the loop.

https://github.com/ros-perception/perception_pcl/blob/0d27ed78ee3a7071a6508516a166bab0640bcba8/pcl_ros/src/pcl_ros/filters/voxel_grid.cpp#L79-L81

SteveMacenski: How about we remove this from this PR and handle separately since its a separate feature request this is adding? At least for me, it would make the diffs easier for me to process without the massive red blocks and other changes.

asymingt: Sure -- I can do that.

pcl_ros/include/pcl_ros/filters/voxel_grid.hpp
asymingt: Can we do this for all filters as a follow-up PR? Otherwise, we'll have one filter with the implementation in the cpp file, whereas all others are declared in the hpp file.

SteveMacenski: OK

asymingt:
@daisukes and @SteveMacenski -- I have reverted the changes corresponding to recent contribution #393 and I'll open up a second PR to add them back after this one merges. Locally, things are passing; as usual CI will surface some unrelated linter errors. Otherwise, I think we're good to go.

SteveMacenski (approved):

PR #402, opened 2023-02-14
Avoid copying data in fromROSMsg with PCL >= 1.13.1

mvieth:
See also https://github.com/PointCloudLibrary/pcl/pull/5608
Supersedes https://github.com/ros-perception/perception_pcl/pull/368

I would add the same changes to melodic-devel, too. Any other branches?


SteveMacenski (approved):

SteveMacenski:
I think melodic would be the other relevant branch

PR #406, opened 2023-02-24
[ros2] lint the pcl_ros code base to try and get a green build :)

asymingt:
Mostly header ordering, but also some uncrustification and a few nitty pep8 errors.


asymingt:
**GREEN GREEN GREEN**

happy

pcl_ros/include/pcl_ros/pcl_node.hpp
```diff
@@ -249,7 +244,9 @@ class PCLNode : public rclcpp::Node
     * \param topic_name an optional topic name (only used for printing, defaults to "indices")
     */
   inline bool
-  isValid(const PointIndicesConstPtr & indices, const std::string & topic_name = "indices")
+  isValid(
+    [[maybe_unused]] const PointIndicesConstPtr & indices,
```
SteveMacenski: For just defining, you can not post a name or comment out via `/*indices*/`. This is typically how codebases I've seen handle this issue - but I didn't actually know about this `maybe_unused` option. This is what I'd prefer if possible, but I won't die on a hill about it

SteveMacenski: These appear to be the only instances, so its not a huge delta change (I probably wouldn't ask you to do it if it required changes 20x+)

SteveMacenski (changes requested):
Thanks so much for this! This makes reviewing PRs much easier for me :-)

SteveMacenski:
I also wish there was an easier gif-generator-poster bot so I could have a bunch of funny keywords result in memes :laughing:

pcl_ros/include/pcl_ros/pcl_node.hpp
asymingt: Ok I replaced the four instances of `[[maybe_unused]]` with commenting out.

SteveMacenski (approved):

PR #419, opened 2023-06-09
Set sensor data QoS to filter subscribers

### LINKED ISSUE #418: ROS2 filter subscriber is not sensor data QoS.
If `use_indices_` is true, subscribers are set sensor data QoS.

https://github.com/ros-perception/perception_pcl/blob/b2dcad2ff15e92542270d490442e2e1207db38cb/pcl_ros/src/pcl_ros/filters/filter.cpp#L121-L125

We have mixed them up, so we are trying to unify them.
I want to set sensor data QoS to the subscriber without indices.

https://github.com/ros-perception/perception_pcl/blob/b2dcad2ff15e92542270d490442e2e1207db38cb/pcl_ros/src/pcl_ros/filters/filter.cpp#L152-L156

Tacha-S:
fix #418 .

### DIFF
--- a/pcl_ros/src/pcl_ros/filters/filter.cpp
+++ b/pcl_ros/src/pcl_ros/filters/filter.cpp
@@ -118,9 +118,7 @@ pcl_ros::Filter::subscribe()
   // If we're supposed to look for PointIndices (indices)
   if (use_indices_) {
     // Subscribe to the input using a filter
-    auto sensor_qos_profile = rclcpp::QoS(
-      rclcpp::KeepLast(max_queue_size_),
-      rmw_qos_profile_sensor_data).get_rmw_qos_profile();
+    auto sensor_qos_profile = rclcpp::SensorDataQoS().keep_last(max_queue_size_);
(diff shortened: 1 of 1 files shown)


pcl_ros/src/pcl_ros/filters/filter.cpp
```diff
@@ -151,7 +151,7 @@ pcl_ros::Filter::subscribe()
     // Subscribe in an old fashion to input only (no filters)
     sub_input_ =
       this->create_subscription<PointCloud2>(
-      "input", max_queue_size_,
+      "input", rclcpp::SensorDataQoS().keep_last(max_queue_size_),
```
SteveMacenski, line 153: Good catch. Can you update the other version to use this same `rclcpp::SensorDataQoS()` format for consistency?

Tacha-S: I updated QoS format in `use_indices` section.

SteveMacenski (approved):

PR #449, opened 2024-05-05
In PCL 1.14.1 and newer, generate smaller point cloud msgs

mvieth:
See also https://github.com/PointCloudLibrary/pcl/pull/5913


SteveMacenski:
ROS 1 isn't being migrated to 24.04 to get 1.14.1+ as the default binaries. I suppose this is backward compatible, but I just wanted to ask you if you think perhaps now is a good time to stop updating ROS 1 branches since we're now officially off OS' that will ever have ROS 1 installs available and supported

mvieth:
You are right, this is not relevant for people who install PCL via apt, but it can be useful for people who install PCL and perception_pcl from source to get the latest features.

SteveMacenski (approved):

PR #457, opened 2024-06-19
Fixed message filter api

ahcorde:
Related with this PR https://github.com/ros2/message_filters/pull/132

### DIFF
--- a/pcl_ros/include/pcl_ros/filters/filter.hpp
+++ b/pcl_ros/include/pcl_ros/filters/filter.hpp
@@ -137,11 +137,17 @@ private:
   /** \brief Pointer to parameters callback handle. */
   OnSetParametersCallbackHandle::SharedPtr callback_handle_;
 
+  typedef message_filters::sync_policies::ApproximateTime<
+      PointCloud2, PointIndices> SyncApproxPolInputIndices;
+  typedef message_filters::Synchronizer<SyncApproxPolInputIndices> SynchronizerApproxInputIndices;
+
+  typedef message_filters::sync_policies::ExactTime<
+      PointCloud2, PointIndices> SyncExactPolInputIndices;
+  typedef message_filters::Synchronizer<SyncExactPolInputIndices> SynchronizerExactInputIndices;
+
   /** \brief Synchronized input, and indices.*/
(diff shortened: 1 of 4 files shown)


ahcorde:
There is an issue with rosdistro and branches. The problem is described here https://github.com/ros-perception/perception_pcl/issues/458

If we merge this PR we will break `humble`, `iron` and `jazzy` deb builds

clalancette:
To the maintainers here: it is usually our policy to not hard-break API like this in the ROS 2 core.  We inadvertently did that when we merged in https://github.com/ros2/message_filters/pull/129 ; we have a revert PR open in https://github.com/ros2/message_filters/pull/132.  However, we do eventually want to move in the direction of deprecating the current API and moving to a new one.

What's your feeling here?  Would it be a lot better for you if we maintained the current API, and then somehow figured out how to move to a new one?

SteveMacenski:
ros2 targets Rolling, humble/iron are branched separately. I don't think this is a problem, no?

clalancette:
> ros2 targets Rolling, humble/iron are branched separately. I don't think this is a problem, no?

It doesn't look like that is the case currently; all of Humble, Iron, Jazzy, and Rolling are using the `ros2` branch:

* https://github.com/ros/rosdistro/blob/08be95e0407cf768d4b37a6e36a99b0d741634c8/humble/distribution.yaml#L5655
* https://github.com/ros/rosdistro/blob/08be95e0407cf768d4b37a6e36a99b0d741634c8/iron/distribution.yaml#L4399
* https://github.com/ros/rosdistro/blob/08be95e0407cf768d4b37a6e36a99b0d741634c8/jazzy/distribution.yaml#L4435
* https://github.com/ros/rosdistro/blob/08be95e0407cf768d4b37a6e36a99b0d741634c8/rolling/distribution.yaml#L4369

SteveMacenski:
See https://github.com/ros-perception/perception_pcl/issues/458

clalancette:
See https://github.com/ros2/message_filters/pull/132#issuecomment-2192225078 , where we reverted the problematic change for now.  We'll probably move back in that direction, so this PR will eventually be needed; up to you whether you want to leave this open for now or close it, and we can reopen later.

PR #459, opened 2024-06-21
pcd_to_pointcloud: add latch param using transient_local QoS

remco-r:
I missed this feature from ROS1. Implemented using transient_local QoS.
Tested in humble using:
`ros2 run pcl_ros pcd_to_pointcloud --ros-args -p file_name:=<file_path> -p latch:=true`

Afterwards a single pointcloud was received with `ros2 topic echo` or in `rviz`, while these nodes were not active when the pointcloud was published.


remco-r:
as nobody cares to comment or review im closing the PR

PR #469, opened 2024-10-29
Deprecating tf2 C Headers

CursedRock17:
Related to this [pull request](https://github.com/ros2/geometry2/pull/720) in `geometry2` in which we deprecated the `.h` style headers in favor of `.hpp`.

### DIFF
--- a/pcl_ros/include/pcl_ros/impl/transforms.hpp
+++ b/pcl_ros/include/pcl_ros/impl/transforms.hpp
@@ -39,11 +39,11 @@
 
 #include <pcl/common/transforms.h>
 #include <pcl_conversions/pcl_conversions.h>
-#include <tf2/convert.h>
-#include <tf2/exceptions.h>
-#include <tf2/LinearMath/Transform.h>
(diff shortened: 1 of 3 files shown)


Rayman:
I think if we want to do this, we should create backwards compatibilty headers that emit compile warnings so users have time to migrate to the new headers.

Let's see first if the PR on geometry2 is going to be merged.

clalancette:
> I think if we want to do this, we should create backwards compatibilty headers that emit compile warnings so users have time to migrate to the new headers.

Yes, that is exactly what https://github.com/ros2/geometry2/pull/720 does.  This PR is just a preemptive one (thanks @CursedRock17 ) to avoid the warnings once we merge that PR in.

CursedRock17:
With the recent backports merged, this PR should be good for review.

SteveMacenski:
Toggling for CI

SteveMacenski:
@CursedRock17 can you open for message filters? That is causing build warnings here since that isn't updated as well (but not blocking)

PR #472, opened 2024-12-05
Update for Rolling file name changes

SteveMacenski:
Fixes some build warnings and allows packages to successfully build with conservative warnings to errors compiler flags, rolling only

### DIFF
--- a/pcl_conversions/include/pcl_conversions/pcl_conversions.h
+++ b/pcl_conversions/include/pcl_conversions/pcl_conversions.h
@@ -41,8 +41,8 @@
 #include <vector>
 
 #include <rclcpp/rclcpp.hpp>
-#include <message_filters/message_event.h>
-#include <message_filters/message_traits.h>
+#include <message_filters/message_event.hpp>
+#include <message_filters/message_traits.hpp>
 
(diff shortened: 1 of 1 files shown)


Timple:
This does not have to be rolling only. The changes are backwards compatible.

Rayman:
> This does not have to be rolling only. The changes are backwards compatible.

Hmm, no? (tested on jazzy)
```
find -name '*message_traits*'
./include/message_filters/message_filters/message_traits.h
```

Timple:
Well bloody... I do remember the discussion that this all should be backported. 

Guess I stand corrected.

AlexandruSopt:
In my ROS2 Humble WS the error appears:
... /pcl_conversions/include/pcl_conversions/pcl_conversions.h:44:10: fatal error: message_filters/message_event.hpp: No such file or directory
   44 | #include <message_filters/message_event.hpp>
I ran all rosdep dependency checks and still can not get rid of error. If the change is rolled back from *.hpp to *.h than everything is peachy.

Bottom line: I do not think this works for ros2 humble deployments. Checked the message_filters repo and the *.hpp versions of the file are available for rolling , but not for humble.

PR #476, opened 2025-01-16
Add lazy feature

Tacha-S:
I have implemented support for the lazy feature, which was previously marked as TODO.

### DIFF
--- a/pcl_ros/src/pcl_ros/filters/filter.cpp
+++ b/pcl_ros/src/pcl_ros/filters/filter.cpp
@@ -178,12 +178,12 @@ pcl_ros::Filter::Filter(std::string node_name, const rclcpp::NodeOptions & optio
 {
   auto pub_options = rclcpp::PublisherOptions();
   pub_options.event_callbacks.matched_callback = [this](rclcpp::MatchedInfo & /*info*/) {
-    if (pub_output_->get_subscription_count() == 0) {
-      unsubscribe();
-    } else {
-      subscribe();
-    }
-  };
+      if (pub_output_->get_subscription_count() == 0) {
+        unsubscribe();
+      } else {
+        subscribe();
+      }
+    };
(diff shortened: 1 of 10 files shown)


SteveMacenski:
Perfectly approve the PR, but CI is failing, so a test might need to be updated: https://build.ros2.org/job/Jpr__perception_pcl__ubuntu_noble_amd64/16/console and we need to have this in the `ros2` branch so all future distributions have it as well (just open a repeated PR in `ros2` once tests pass and we can merge the pair)

Tacha-S:
I was unsure whether it would be better to call `createPublishers()` in the derived class or to only address the issue with `project_inliers`. I would appreciate any opinions or suggestions on this.

Although I have passed all tests, it seems that the CI is not successful. 
Additionally, due to changes in `message_filters` in #477 (https://github.com/ros-perception/perception_pcl/pull/472), the CI fails.

pcl_ros/src/pcl_ros/filters/filter.cpp
```diff
@@ -176,10 +176,24 @@ pcl_ros::Filter::unsubscribe()
...
+//////////////////////////////////////////////////////////////////////////////////////////////
+void
+pcl_ros::Filter::createPublishers()
+{
+  auto pub_options = rclcpp::PublisherOptions();
+  pub_options.event_callbacks.matched_callback = [this](rclcpp::MatchedInfo & /*info*/) {
+      if (pub_output_->get_subscription_count() == 0) {
+        unsubscribe();
+      } else {
+        subscribe();
```
SteveMacenski: Shouldn't we only subscribe if we previously didn't subscribe? I think we should check the subsciber's state and then subscribe if not processing already

Tacha-S: I modified it to check whether the subscriber is nullptr.

SteveMacenski (approved):
Thank you much!

PR #477, opened 2025-01-17
Add lazy feature to ros2 branch

Tacha-S:
I have applied the same changes as #476 to the ROS 2 branch.

### DIFF
--- a/pcl_ros/src/pcl_ros/filters/filter.cpp
+++ b/pcl_ros/src/pcl_ros/filters/filter.cpp
@@ -176,6 +176,13 @@ pcl_ros::Filter::unsubscribe()
 pcl_ros::Filter::Filter(std::string node_name, const rclcpp::NodeOptions & options)
 : PCLNode(node_name, options)
 {
+  RCLCPP_DEBUG(this->get_logger(), "Node successfully created.");
+}
+
+//////////////////////////////////////////////////////////////////////////////////////////////
+void
+pcl_ros::Filter::createPublishers()
+{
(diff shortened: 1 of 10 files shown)


pcl_ros/src/pcl_ros/filters/filter.cpp
```diff
@@ -176,10 +176,24 @@ pcl_ros::Filter::unsubscribe()
...
+//////////////////////////////////////////////////////////////////////////////////////////////
+void
+pcl_ros::Filter::createPublishers()
+{
+  auto pub_options = rclcpp::PublisherOptions();
+  pub_options.event_callbacks.matched_callback = [this](rclcpp::MatchedInfo & /*info*/) {
+      if (pub_output_->get_subscription_count() == 0) {
+        unsubscribe();
+      } else {
+        subscribe();
```
SteveMacenski: See jazzy PR branch

SteveMacenski (changes requested):

pcl_ros/src/pcl_ros/filters/filter.cpp
Tacha-S: I checked the PR for the jazzy branch, but it seems there are no reviews. Could it still be pending?

SteveMacenski: Sigh. GitHub must have blipped. Its up :-)

SteveMacenski (approved):
Thank you!

PR #481, opened 2025-03-06
Fix pcd timestamp format output in pointcloud_to_pcd

Adraub:
Fix pcd timestamp indentation offset if nanosec value is less than 100000000 ns

### DIFF
--- a/pcl_ros/tools/pointcloud_to_pcd.cpp
+++ b/pcl_ros/tools/pointcloud_to_pcd.cpp
@@ -121,8 +121,9 @@ public:
     }
 
     std::stringstream ss;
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Please save a file using this change and report the file name to validate (since CI doesn't cover the name of file testing).

SteveMacenski (approved):

Adraub:
@SteveMacenski here is the output:
Before:
`[INFO] [1741286117.246267894] [pointcloud_to_pcd]: Writing to 899.11149430.pcd`
After:
`[INFO] [1741286117.246267894] [pointcloud_to_pcd]: Writing to 899.011149430.pcd`

PR #486, opened 2025-04-20
Port bag_to_pcd to ROS 2

vladimirjendrol:
Usage example:
`ros2 run pcl_ros bag_to_pcd --ros-args -p bag_path:=rosbag2_2022_04_14-ped_vehicle/ -p topic_name:=/sensing/lidar/top/rectified/pointcloud -p output_directory:=pcds`

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -293,7 +293,7 @@ install(
 #    pcl_ros_segmentation
     pointcloud_to_pcd_lib
     combined_pointcloud_to_pcd_lib
-    bag_to_pcd
+    bag_to_pcd_lib
 #    convert_pcd_to_image
 #    convert_pointcloud_to_image
   EXPORT export_pcl_ros
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -23,6 +23,7 @@ find_package(tf2 REQUIRED)
 find_package(tf2_eigen REQUIRED)
 find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
+find_package(rosbag2_transport REQUIRED)
 
 set(dependencies
   pcl_conversions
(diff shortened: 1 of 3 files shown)


pcl_ros/tools/bag_to_pcd.cpp
```diff
@@ -44,116 +44,77 @@ Cloud Data) format.
...
-        cloud_t = *cloud;
-      }
+      rclcpp::SerializedMessage serialized_msg(*msg->serialized_data);
+      sensor_msgs::msg::PointCloud2 pointcloud_msg;
+      serialization_.deserialize_message(&serialized_msg, &pointcloud_msg);
 
-      std::cerr << "Got " << cloud_t.width * cloud_t.height << " data points in frame " <<
-        cloud_t.header.frame_id << " with the following fields: " << pcl::getFieldsList(cloud_t) <<
-        std::endl;
+      pcl::PointCloud<pcl::PointXYZ> cloud;
```
Rayman: Right now you are converting from sensor_msgs to `pcl::PointCloud<pcl::PointXYZ>`. This will throw away all extra fields. If you convert to `pcl::PointCloud2` and serialize, all those fields will go into the pcd file.

Rayman (changes requested):
Thank you for your contribution!

Can you fix cpplint? I'm seeing a couple of errors in the CI.

Also a question: why did you change the commandline parsing? Previously it would print usage when you entered the wrong amount of arguments.

Rayman:
Can you resolve the conflicts and implement the conversion with `pcl::PointCloud2` (so all fields are saved). Then we can merge this I think

pcl_ros/CMakeLists.txt
```diff
@@ -214,6 +215,21 @@ ament_target_dependencies(combined_pointcloud_to_pcd_lib
...
+ament_target_dependencies(bag_to_pcd_lib
+  rclcpp
+  rclcpp_components
+  sensor_msgs
+  pcl_conversions
+  rosbag2_transport)
+rclcpp_components_register_node(bag_to_pcd_lib
+  PLUGIN "pcl_ros::BagToPCD"
+  EXECUTABLE bag_to_pcd)
+
```
Rayman, line 232: Can you make sure to also install `bag_to_pcd_lib`? I'm getting the following error:
```
ros2 run pcl_ros bag_to_pcd 
terminate called after throwing an instance of 'class_loader::LibraryLoadException'
  what():  Could not load library dlopen error: libbag_to_pcd_lib.so: cannot open shared object file: No such file or directory, at ./src/shared_library.c:99
[ros2run]: Aborte
```

Rayman:
Thank you for this feature!

PR #488, opened 2025-04-29
Added crop box marker publisher

alireza-moayyedi:
Hi,

I always found it a bit frustrating not being able to visualize the crop box specially when things don't go as expected and I would need to debug what is going wrong.

I thought it would be nice to have a marker msg which correctly (taking into account the crop box frame) visualizes the crop box filter. 

Screenshot from 2025-04-29 16-57-00
Screenshot from 2025-04-29 16-56-48

The publishing logic could  be a discussion point. I thought that it would be best to prepare the box characteristics upon parameter changes and publish it with the pointcloud cb. I also tried alternative approaches such as publishing a transient local message with parameter updates but that did not seem so practical because of the following reasons:

- Not necessarily, input_frame is set.
- The incoming messages can have different header frames, hence the marker should always use the latest one.
- Pointclouds can come after the initial parameter configuration, hence you would need to implement flags to retrigger marker publishing.
- etc.

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -23,6 +23,7 @@ find_package(tf2 REQUIRED)
 find_package(tf2_eigen REQUIRED)
 find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
+find_package(visualization_msgs REQUIRED)
 
 set(dependencies
   pcl_conversions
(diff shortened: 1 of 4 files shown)


SteveMacenski (approved):
@Rayman ?

Rayman (approved):
Thank you for this feature! Can you rebase to fix the cpplint errors?

Personally I wouldn't have cached the visualization message in a member, because the operations there are very cheap and happen on the stack (copying of floats). But I'll accept this also.

PR #489, opened 2025-04-29
Modern CMake

### LINKED ISSUE #487: Modern CMake for `pcl_conversions`
`pcl_conversions` uses old CMake and Ament idioms.

A target should be defined, of type INTERFACE, and properly exported.

roncapat:
Closes #487

### DIFF
--- a/pcl_conversions/CMakeLists.txt
+++ b/pcl_conversions/CMakeLists.txt
@@ -33,6 +33,7 @@ add_library(${PROJECT_NAME} INTERFACE)
 target_include_directories(
   ${PROJECT_NAME}
   INTERFACE
+  "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
   "$<INSTALL_INTERFACE:include/${PROJECT_NAME}>"
 )
 
@@ -63,7 +64,7 @@ if(BUILD_TESTING)
   find_package(ament_cmake_gtest REQUIRED)
 
   ament_add_gtest(${PROJECT_NAME}-test test/test_pcl_conversions.cpp)
-  target_link_libraries(${PROJECT_NAME}-test ${PROJECT_NAME} ${Boost_LIBRARIES})
+  target_link_libraries(${PROJECT_NAME}-test ${PROJECT_NAME})
 endif()
 
 ament_export_targets(${PROJECT_NAME}_export)
--- a/pcl_conversions/CMakeLists.txt
+++ b/pcl_conversions/CMakeLists.txt
@@ -28,13 +28,34 @@ set(dependencies
   std_msgs
 )
 
-include_directories(
+add_library(${PROJECT_NAME} INTERFACE)
+
+target_include_directories(
+  ${PROJECT_NAME} INTERFACE
   include
   ${PCL_COMMON_INCLUDE_DIRS}
 )
 
-install(DIRECTORY include/
+target_link_libraries(
+  ${PROJECT_NAME} INTERFACE
+  rclcpp::rclcpp
+  message_filters::message_filters
+  pcl_common
+  pcl_io
+  ${std_msgs_TARGETS}
+  ${pcl_msgs_TARGETS}
+  ${sensor_msgs_TARGETS}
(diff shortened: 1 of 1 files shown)


roncapat:
Oops, will fix these simple CI faults tomorrow :)

SteveMacenski (approved):
@Rayman any objections (once CI passes)?

roncapat:
How CI works in this repo? The re-triggering is automatic or not?

alireza-moayyedi:
Hi, recently I also had to do somehting similar for one of  my own packages with PCL and Boost. I wonder if you also considered these:

- Looking at the cmake config of the pcl library, you see that they indicate using target_link_libraries only (hence `${PCL_COMMON_INCLUDE_DIRS}` is not needed):
    - `cat /usr/lib/x86_64-linux-gnu/cmake/pcl/PCLConfig.cmake` shows the following:
Screenshot from 2025-04-30 11-44-49.
In other words, `target_link_libraries(<your_target> ...${PCL_COMMON_LIBRARIES})` for instance would work fine as  `${PCL_COMMON_LIBRARIES}` seems to be a target by itself hence no need to specify its include dirs.

- You could also try getting only the required components from boost. For instance:
    - `find_package(Boost REQUIRED COMPONENTS headers)`
    - `target_link_libraries(<your_target> Boost::headers)`

I may be wrong, but just wanted to share with you what worked for me.

Rayman (approved):
Thank you!

Rayman:
Can you fix the CI failure?
```
08:59:55 --- stderr: pcl_conversions
08:59:55 CMake Error at CMakeLists.txt:33 (target_include_directories):
08:59:55   target_include_directories may only set INTERFACE properties on INTERFACE
08:59:55   targets
```

roncapat:
Sure, in a few hours. About the CI, do you know why it isn't triggered automatically? Does toggle-closing & reopening the PR works?

Edit: it seems the last run was about 4h ago, in correspondence to my commit, so I believe CI gets triggered by "running" status (yellow) just doesn't get reported here on GH.

roncapat:
In particular, I would like to select only the exact subset of PCL targets needed by this package, and check also the suggestions of @alireza-moayyedi (that are somewhat linked, in particular the one on PCL).

roncapat:
BTW I think boost is not required at all... will give it a try after this CI run

roncapat:
@Rayman done! Humble job failure seems unrelated, am I right?

alireza-moayyedi:
> BTW I think boost is not required at all... will give it a try after this CI run

Hi @roncapat , There are Boost related header files being included in the code. You may be able to compile without explicitly  linking against the boost libraries, but that does not mean you should do so. Simply put, this would be then a masked problem. Boost is used by many different packages. Hence, there is a high chance one of the other imported packages has already linked against it so cmake manages to compile without throwing an error. Same thing can happen with rclcpp for instance. There is a high chance some other package which you import already links against it hence one may be able to compile their node without explicitly linking. So in short, even though you can compile, I suggest being explicit in everything that your targets require to link against.

roncapat:
@alireza-moayyedi Well, it seems to me that Boost was only cmake-linked in the test target, so it won't affect downstream IMHO. The "library" target doesn't use it at all, and nobody will ever link to the test target on the other side.

alireza-moayyedi:
> @alireza-moayyedi Well, it seems to me that Boost was only cmake-linked in the test target, so it won't affect downstream IMHO. The "library" target doesn't use it at all, and nobody will ever link to the test target on the other side.

@roncapat You are indeed correct and I apologize for my confusion. I was looking at pcl_ros whereas you are focusing on pcl_conversions. Though I see that boost is a matter of fact missing from pcl_ros cmake. But that's a separate issue. Again, sorry for my confusion.

roncapat:
@alireza-moayyedi no worries at all :) I always appreciate fast iterations and reviews.

Rayman (approved):

Rayman:
humble CI is unrelated and should be fixed by https://github.com/ros/rosdistro/pull/45568

PR #490, opened 2025-04-30
Fix CI: properly order the header includes for cpplint

alireza-moayyedi:
Noticed this error in other PRs so I decided to fix it.


Rayman (approved):

Rayman:
Thank you for the PR. The CI on humble fails, but this is expected. I've tried to fix it here: https://github.com/ros/rosdistro/pull/45568

PR #491, opened 2025-05-01
Improve pointcloud_to_pcd by extracting all fields available in sensor_msgs PointCloud2

Adraub:
Before pointcloud_to_pcd tool was not able to extract intensities out of a sensor_msgs::msg::PointCloud2 message.
Now it can extract automatically any field like XYZ, XYZI, XYZRGB, etc...
I have also done a bit of refactoring to use pcl_ros::transformPointCloud API rather than doing computation in the tool code.


Rayman (approved):

Rayman:
Thank you!

PR #506, opened 2025-08-18
chore: tf2_ros to hpp headers

Timple:


pcl_ros/include/pcl_ros/impl/transforms.hpp
```diff
@@ -39,8 +39,8 @@
 
 #include <pcl/common/transforms.h>
 #include <pcl_conversions/pcl_conversions.h>
-#include <tf2_ros/buffer.h>
-#include <tf2_ros/transform_listener.h>
+#include <tf2_ros/buffer.hpp>
+#include <tf2_ros/transform_listener.hpp>
```
Rayman: I think cpplint now fails because now the .hpp include should go with the other .hpp includes

Timple: Hmm, guess you're right. If only we had tools with a `--reformat` option :smile:

PR #505, opened 2025-08-12
feat(humble): Migrated most of the source and include codes from ROS1 to ROS2

diogovazguerra:
I migrated most of files from ROS1 to ROS2 of pcl_ros package

PR #501, opened 2025-06-04
Enable filters in humble

barryridge:
Related thread: #423

I don't know if you folks accept PRs for previously released distros, i.e. `humble` in this case?

We've been doing some work on Jetson Orin Nano boards, which have not yet been bumped to `noble`/`jazzy` yet AFAIK, and needed to use the PCL filters, so I thought it would be a good idea to try to enable them in the `pcl_ros` `humble` build.

The general strategy I used was to cherry-pick the requisite filter updates from `jazzy` and merge them into `humble`.  This required some slight modifications detailed as follows.

I modified `pcl_ros::Filter::subscribe()` to use `humble`-compatible `message_filters::Subscriber::subscribe()` calls, changing this from `jazzy`:

```c++
    // Subscribe to the input using a filter
    auto sensor_qos_profile =
      rclcpp::SensorDataQoS().keep_last(max_queue_size_).get_rmw_qos_profile();
    sub_input_filter_.subscribe(this, "input", sensor_qos_profile);
    sub_indices_filter_.subscribe(this, "indices", sensor_qos_profile);
```

to this:

```c++
    // Subscribe to the input using a filter
    auto sensor_qos_profile = rclcpp::SensorDataQoS().keep_last(max_queue_size_);
    sub_input_filter_.subscribe(this, "input", sensor_qos_profile.get_rmw_qos_profile());
    sub_indices_filter_.subscribe(this, "indices", sensor_qos_profile.get_rmw_qos_profile());
```

I also modified `pcl_ros::Filter::createPublishers()` to avoid the use of missing `rclcpp::PublisherOptions.event_callbacks.matched_callback`, swapping this in `jazzy`:

```c++
void
pcl_ros::Filter::createPublishers()
{
  auto pub_options = rclcpp::PublisherOptions();
  pub_options.event_callbacks.matched_callback = [this](rclcpp::MatchedInfo & /*info*/) {
      if (pub_output_->get_subscription_count() == 0) {
        unsubscribe();
      } else {
        if (use_indices_) {
          if (!sub_input_filter_.getSubscriber() || !sub_indices_filter_.getSubscriber()) {
            subscribe();
          }
        } else {
          if (!sub_input_) {
            subscribe();
          }
        }
      }
    };
  pub_output_ = create_publisher<PointCloud2>("output", max_queue_size_, pub_options);
}
```

for this:

```c++
void
pcl_ros::Filter::createPublishers()
{
  auto pub_options = rclcpp::PublisherOptions();
  if (use_indices_) {
[...]

Issue #423, opened 2023-07-06
pcl-ros include error under ros2 humble enviornment


lightinfection:
Hi, I install ros-humble-pcl-ros and am trying to include the "point_cloud.hpp" header file for developing a project, but here's an error about the include file in it. I find the ros.h is sourced on the 45th row, but this should not be supported in ros2. May I know if it's a bug or if I misunderstand anything here? Thank you.

#include <ros/ros.h>
#include <pcl/point_cloud.h>
#include <pcl/point_traits.h>
#include <pcl/for_each_type.h>
#include <pcl/conversions.h>

mvieth:
Fixed on branch `ros2` (https://github.com/ros-perception/perception_pcl/commit/3bb07c07c84e67df25f7001ea7022b0b40149059)

felixf4xu:
@wkentaro I see your name on the maintainer list, could you help to close this issue?

this repo has too many _open_ issues but some of them could be closed.

jonselling:
These changes were made on the `ros2` branch, is it possible to backport these changes to humble so the apt packages get updated?

jonselling:
@SteveMacenski Pinging since you are a maintainer in pcl_ros as well, is there a way we can backport this fix to `humble`? It looks like it was this PR: https://github.com/ros-perception/perception_pcl/pull/410

andrewjong:
Bump, just tried to install ros2-humble-pcl-ros and it errored with ROS1. Can we update it in the Ubuntu package repository please? Thanks!

SteveMacenski:
Please cherry pick it back to `humble` and I can merge + run a release. I don't see that that commit is in the humble branch at all for a new release to change anything.

SteveMacenski:
Merged, will do release this afternoon :-)

andrewjong:
Hurray! Thank you so much!

SteveMacenski:
https://github.com/ros/rosdistro/pull/42460

AbhiKhoyani:
[...]

PR #507, opened 2025-08-26
Humble Porting

AntoBrandi:
Replaces https://github.com/ros-perception/perception_pcl/pull/505 and https://github.com/ros-perception/perception_pcl/pull/501 by properly cherry-picking `ros2` commits into `humble`.

Addresses: https://github.com/ros-perception/perception_pcl/issues/423

Additionally, to fix the build on Humble, small fixes were required:

- QoS profiles to use `.get_rmw_qos_profile()` in subscribers
- Disable lazy subscribers as `pub_options.event_callbacks.matched_callback = [this](rclcpp::MatchedInfo & /*info*/) ...` is not available in Humble


AntoBrandi:
A note on my last commit: I can try to make it work on different ROS distros so that it can be cherry-picked in the master. I can use the same approach used in the [laser_filters](https://github.com/ros-perception/laser_filters/blob/rolling/src/scan_to_scan_filter_chain.cpp#L92) package.
Alternatively, the simpler [ros2_control](https://github.com/ros-controls/ros2_control/blob/master/controller_manager/src/controller_manager.cpp#L47) approach can also be used.

This way, future syncs should be easier

Rayman:
Thank you for putting the effort to cherry-pick all relevant commits to humble

I have questions about two specific commits:

1. `Modern CMake`: This changes the exported targets. Will this cause downstream compilation failures?
2. `Fix warning: 'subscribe<>' is deprecated`: This commit is only needed for jazzy and higher. You corrected this in a lter commit but it might be better to just remove this commit from this PR so you don't need to correct it
3. In your last commit you did a linter fix. Does it make more sense to disable uncrustify so we can freely cherry-pick from the `ros2` branch?

In general I'm a bit worries the changes will lead to downstream compilation failures. Can @SteveMacenski maybe give his opinion about this PR?

Btw, I'm going on vacation for 3 weeks so it might take a while before I come back to this.

AntoBrandi:
My original idea was to minimize the differences between the `ros2` branch and the `humble` branch, so that future commits could be easily backported to `humble`.
However, I understand your point, especially for the possible downstream compilation failures.
In that case, we can follow a different approach: Making the `ros2` branch buildable on `humble`, like the [`humble_main` branch in Nav2](https://github.com/ros-navigation/navigation2/tree/humble_main)

SteveMacenski:
I wouldn't change the exported targets / variables in an already released distribution. Rclcpp I see still uses the old ament macro for this in Humble https://github.com/ros2/rclcpp/blob/humble/rclcpp/CMakeLists.txt#L193-L207 so that makes me further think we shouldn't backport that to humble as a ABI breaking change. 

P.S. I don't particularly recommend the `humble_main` strategy, its a pain in the ass to keep synced up and I've actually already since decided to stop after introducing `nav2_ros_common` due to extensive use of ROS 2 advanced features not available in Humble at all. It won't be updated post-July 2025 (though up to date in a broad sense compared to humble itself). The changes for targets + APIs makes keeping Humble x Rolling unrealistically difficult to maintain compatibility for in a project unless someone wants to take explicit ownership for it as a prime objective with regular care and upkeep.

Issue #225, opened 2019-03-27
ROS2 Port


LucidOne:
Is there a central place where we are tracking architectural  issues for a ROS2 interface to PCL?

The big one I recall is handling pipelined pcl operations with nodelets. Zero copy is great until an old point cloud gets over written by a new one before you are done with it.

paulbovbel:
Not really, as no one has declared they have time/inclination to make it happen. The way I see it, the two missing pieces required for a nominally useful ROS2 PCL pipeline are:
- type masquerading - serializing/deserializing between every pipeline step is unpleasant
- a way to compose and execute a pipeline from a bunch of managed nodes. I suspect most of the pieces are there (other than perhaps a pipeline-aware executor), but documentation is a bit sparse at the moment.

> The big one I recall is handling pipelined pcl operations with nodelets. Zero copy is great until an old point cloud gets over written by a new one before you are done with it.

I imagine it's the middleware's job to handle object lifetime, and intraprocess comms in ROS2 works by more-or-less passing a dummy message with a memory address. So there's probably an internal mechanism that keeps the messages around until all const smart_pointers have expired.

---

EDIT, well not no-one, but no-one currently. https://discourse.ros.org/t/ros2-pcl-perceptions/3596/23

LucidOne:
I'll jot down some of my other thoughts as I remember them here until there is a designated place to keep track of design notes.

I'd like to consider adding time delta information to ordered point clouds so the temporal dewarping code doesn't need to be reimplemented for each lidar. My recollection is that the rotation speed of a spinning lidar was slightly susceptible to power supply fluctuations but the PLL did a fairly good job besides that. This may be more or less of an issue in the future with solid state lidars.

I think there are some potential algorithmic benefits to to considering the projective origin of a pointcloud. Along this line of thinking, it may be worth evaluating non-cartesian error modeling as many of the current lidar designs have anisotropic error. As I understand it, a Velodyne HDL-32 for example has a maximum measurement error along the beam path, fixed angular offset error between the set of diodes epoxied to the PCB, minimal gyro wobble and minimal encoder error around the axis of rotation. Projective origin information can also allow diagnostics to detect hardware failures, an encoder failure in a rotating lidar might be detected by a change in the bin counts of an angular histogram from the projective origin..

Some of my conversations with lidar manufacturers seem to indicate that solid state lidar are capable of beam steering so it may be worth considering what might be needed to support a variable angular resolution.

[...]

PR #510, opened 2025-09-02
Migrate pcl_ros::SACSegmentation to ROS 2

AntoBrandi:
Init the migration of some `segmentation` components with the `pcl_ros::SACSegmentation`.

Relates to https://github.com/ros-perception/perception_pcl/issues/225

Future Improvements:
Given the quantity of boilerplate code that I had to duplicate from [`filter.hpp`](https://github.com/ros-perception/perception_pcl/blob/ros2/pcl_ros/include/pcl_ros/filters/filter.hpp) and that will be required in other segmentation Components as well, what about including it in the [`pcl_node.hpp`](https://github.com/ros-perception/perception_pcl/blob/ros2/pcl_ros/include/pcl_ros/pcl_node.hpp) ?
I refer to members like:
- `use_frame_params()`
- `input_indices_callback()`
- `createPublishers()`
- `unsubscribe()`
- `subscribe()`
- `computePublish()`

### DIFF
--- a/pcl_ros/src/pcl_ros/segmentation/sac_segmentation.cpp
+++ b/pcl_ros/src/pcl_ros/segmentation/sac_segmentation.cpp
@@ -250,8 +250,8 @@ pcl_ros::SACSegmentation::subscribe()
   if (use_indices_) {
     // Subscribe to the input using a filter
     auto sensor_qos_profile = rclcpp::SensorDataQoS().keep_last(max_queue_size_);
-    sub_input_filter_.subscribe(this, "input", sensor_qos_profile.get_rmw_qos_profile());
-    sub_indices_filter_.subscribe(this, "indices", sensor_qos_profile.get_rmw_qos_profile());
+    sub_input_filter_.subscribe(this, "input", sensor_qos_profile);
+    sub_indices_filter_.subscribe(this, "indices", sensor_qos_profile);
 
     if (approximate_sync_) {
       sync_input_indices_a_ =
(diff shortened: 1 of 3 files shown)


pcl_ros/src/pcl_ros/segmentation/sac_segmentation.cpp
```diff
@@ -135,56 +249,39 @@ pcl_ros::SACSegmentation::subscribe()
...
-      // Subscribe to a callback that saves the indices
-      sub_indices_filter_.registerCallback(bind(&SACSegmentation::indices_callback, this, _1));
-      // Subscribe to a callback that sets the header of the saved indices to the cloud header
-      sub_input_filter_.registerCallback(bind(&SACSegmentation::input_callback, this, _1));
-
-      // Synchronize the two topics. No need for an approximate synchronizer here, as we'll
-      // match the timestamps exactly
+    auto sensor_qos_profile = rclcpp::SensorDataQoS().keep_last(max_queue_size_);
+    sub_input_filter_.subscribe(this, "input", sensor_qos_profile.get_rmw_qos_profile());
+    sub_indices_filter_.subscribe(this, "indices", sensor_qos_profile.get_rmw_qos_profile());
```
Rayman: In rolling they removed the api method that uses a `rmw_qos_profile_t`. So you can just directly pass the `rclcpp::Qos` object to subscribe.

AntoBrandi: right! I removed it

PR #518, opened 2026-02-09
Add keep organized parameter

### LINKED ISSUE #517: Add `keep_organized` option to more pcl_ros filters
In pcl_ros, only a few filters such as `CropBox` and `PassThrough` expose the `keep_organized` option.

Filters like `RadiusOutlierRemoval` and `StatisticalOutlierRemoval` do not expose this option, even though the underlying PCL filters support `keep_organized`. 

Since it is technically possible, I would like to add this option to these filters.

Tacha-S:
I have also added the `keep_organized` parameter to the following filters:
- extract_indices
- radius_outlier_removal
- statistical_outlier_removal

fix #517

### DIFF
--- a/pcl_ros/src/pcl_ros/filters/extract_indices.cpp
+++ b/pcl_ros/src/pcl_ros/filters/extract_indices.cpp
@@ -46,12 +46,20 @@ pcl_ros::ExtractIndices::ExtractIndices(const rclcpp::NodeOptions & options)
   neg_desc.description = "Extract indices or the negative (all-indices)";
   declare_parameter(neg_desc.name, rclcpp::ParameterValue(false), neg_desc);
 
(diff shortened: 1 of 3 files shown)


SteveMacenski (commented):
Is this a common param perhaps across all filters? Then adding to the base class may make more sense and then use it in the derived classes as needed https://github.com/ros-perception/perception_pcl/blob/ros2/pcl_ros/src/pcl_ros/filters/filter.cpp#L240

Tacha-S:
Filters that inherit from `pcl::FilterIndices` provide the `keep_organized` option, whereas filters inheriting from `pcl::Filter` do not, such as `ProjectInliers` and `VoxelGrid`.

SteveMacenski (approved):

PR #519, opened 2026-02-10
pcl_ros: enable QoS overrides for lazy input and ouput

Spike-hd:
Currently, the input subscription uses a fixed QoS profile (SensorDataQoS) and does not support ROS 2 QoS parameter overrides. This prevents users from customizing QoS for compatibility with upstream publishers and subscribers.

### DIFF
--- a/pcl_ros/src/pcl_ros/filters/filter.cpp
+++ b/pcl_ros/src/pcl_ros/filters/filter.cpp
@@ -37,6 +37,7 @@
 
 #include "pcl_ros/filters/filter.hpp"
 #include <pcl/common/io.h>
+#include <rclcpp/qos_overriding_options.hpp>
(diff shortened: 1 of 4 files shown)


SteveMacenski (commented):
Please lint, but otherwise LGTM. Are there other places we should add too?

SteveMacenski (approved):

SteveMacenski:
@Spike-hd please open for Rolling and I can merge for both rolling and jazzy (so once you move to Lyrical, it'll be ready for you there too)

Spike-hd:
Hi, thanks for reviewing the previous PR. As requested, I’ve now created a version of this patch targeting Rolling (ros2 branch):
https://github.com/ros-perception/perception_pcl/pull/521

PR #520, opened 2026-02-12
[Backport Jazzy]Add keep organized parameter (#518 )

Tacha-S:
Backport #518 to Jazzy branch.

### DIFF
--- a/pcl_ros/src/pcl_ros/filters/extract_indices.cpp
+++ b/pcl_ros/src/pcl_ros/filters/extract_indices.cpp
@@ -46,12 +46,20 @@ pcl_ros::ExtractIndices::ExtractIndices(const rclcpp::NodeOptions & options)
   neg_desc.description = "Extract indices or the negative (all-indices)";
   declare_parameter(neg_desc.name, rclcpp::ParameterValue(false), neg_desc);
 
+  rcl_interfaces::msg::ParameterDescriptor keep_organized_desc;
+  keep_organized_desc.name = "keep_organized";
+  keep_organized_desc.type = rcl_interfaces::msg::ParameterType::PARAMETER_BOOL;
+  keep_organized_desc.description =
+    "Set whether the filtered points should be kept and set to NaN, "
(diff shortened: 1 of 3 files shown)


SteveMacenski (commented):
Since this is a released distro, I just want to make sure: are the defaults for all of these `false` when not set prior to now -- so there is no behavioral change?

Tacha-S:
Of course.

According to the [PCL documentation](https://pointclouds.org/documentation/classpcl_1_1_filter_indices.html#a45e39a4237dfff1735294cacc243a918), the default behavior is to output a dense point cloud.
We have also confirmed this through actual behavior.

For example, with a 640×480 input point cloud:

```
header:
  stamp:
    sec: 206
    nanosec: 780000000
  frame_id: head_camera_rgb_optical_frame
height: 480
width: 640
fields: '<sequence type: sensor_msgs/msg/PointField, length: 4>'
is_bigendian: false
point_step: 32
row_step: 20480
data: '<sequence type: uint8, length: 9830400>'
is_dense: false
```

When running:

```
ros2 run pcl_ros filter_statistical_outlier_removal_node \
  --ros-args -r input:=/camera/depth_registered/points -p stddev:=1.0
```

the output becomes:

```
header:
  stamp:
    sec: 206
    nanosec: 780000000
  frame_id: head_camera_rgb_optical_frame
height: 1
width: 278905
fields: '<sequence type: sensor_msgs/msg/PointField, length: 4>'
is_bigendian: false
point_step: 32
row_step: 8924960
data: '<sequence type: uint8, length: 8924960>'
is_dense: false
```

Rayman:
I'm going to look at the difference between the `ros2` branch and `jazzy` and see if I can backport some more non-breaking changes

Rayman:
I've collected all commits that could be backported, including yours, in this PR: https://github.com/ros-perception/perception_pcl/pull/522

Please take a look at it. I'll run some more test to see if it would not break anything.

PR #521, opened 2026-02-12
pcl_ros: apply QoS overrides to filters and tools

Spike-hd:
Currently, the input subscription uses a fixed QoS profile (SensorDataQoS) and does not support ROS 2 QoS parameter overrides. This prevents users from customizing QoS for compatibility with upstream publishers and subscribers.
This PR applies the QoS overrides patch to pcl_ros filters and tools.

### DIFF
--- a/pcl_ros/src/pcl_ros/filters/filter.cpp
+++ b/pcl_ros/src/pcl_ros/filters/filter.cpp
@@ -134,8 +134,8 @@ pcl_ros::Filter::subscribe()
   if (use_indices_) {
     // Subscribe to the input using a filter
     auto sensor_qos_profile = rclcpp::SensorDataQoS().keep_last(max_queue_size_);
-    sub_input_filter_.subscribe(this, "input", sensor_qos_profile);
-    sub_indices_filter_.subscribe(this, "indices", sensor_qos_profile);
+    sub_input_filter_.subscribe(this, "input", sensor_qos_profile, sub_options);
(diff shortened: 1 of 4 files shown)


SteveMacenski (approved):

SteveMacenski:
^ Check CI output please

pcl_ros/src/pcl_ros/filters/filter.cpp
```diff
@@ -120,6 +120,16 @@ pcl_ros::Filter::computePublish(
...
+  // Enable QoS reconfigurability via parameters
+  rclcpp::SubscriptionOptions sub_options;
+  sub_options.qos_overriding_options =
+    rclcpp::QosOverridingOptions {{
+    rclcpp::QosPolicyKind::History,
+    rclcpp::QosPolicyKind::Reliability,
+    rclcpp::QosPolicyKind::Durability,
+    rclcpp::QosPolicyKind::Depth
+  }};
+
```
Rayman, line 132: I don't see this variable used anywhere

Rayman (changes requested):
PR looks good, I have one small comment

pcl_ros/src/pcl_ros/filters/filter.cpp
Spike-hd: You are right thanks, I just fixed it in the lastest commit.

Rayman (approved):

PR #528, opened 2026-03-24
fix(pcl_ros): preserve all fields in combined_pointcloud_to_pcd

kubojion:
`combined_pointcloud_to_pcd` was accumulating clouds into typed
`pcl::PointCloud<pcl::PointXYZ>` / `pcl::PointCloud<pcl::PointXYZRGB>`
containers, silently dropping all extra fields (intensity, ring,
timestamps, custom fields, etc.) on every incoming message.

This applies the same approach used in `pointcloud_to_pcd` since #491:
replace the typed accumulators with a single `pcl::PCLPointCloud2` so
all fields are preserved as-is.

The TF transform path also needed to change: typed
clouds allowed manual Eigen loops over `.points`, which is not possible
with `PCLPointCloud2`. The fix uses `pcl_ros::transformPointCloud` (same
as `pointcloud_to_pcd`). With per-cloud TF, failing silently and falling
back to the original frame would mix coordinate frames in the output, so
the node now skips the cloud and logs a warning instead.

Other changes:
- `pcl::PCLPointCloud2::concatenate` replaces manual buffer append,
  adding schema validation (skips cloudds with mismatched fields or endianness)
- `rgb` parameter is kept but marked deprecated
- `is_dense` propagated as logical AND across accumulated clouds
- `pcl_ros_tf` added to link dependencies; `tf2_eigen` and direct
  `tf2_ros` removed (transitively provided by `pcl_ros_tf`)
- Fixed filename bug: prefix `"combined_"` was producing
  `combined_combined_*.pcd`
- `rclcpp::shutdown()` no longer called from destructor (passed via
  `do_shutdown` flag to avoid UB)

### DIFF
--- a/pcl_ros/tools/combined_pointcloud_to_pcd.cpp
+++ b/pcl_ros/tools/combined_pointcloud_to_pcd.cpp
@@ -130,7 +130,7 @@ public:
     // Optionally save on node shutdown (do not call rclcpp::shutdown from destructor)
     if (!save_triggered_ && save_on_shutdown_) {
       RCLCPP_INFO(this->get_logger(), "Node is shutting down; saving accumulated cloud.");
-      saveAccumulatedCloud(false);
+      saveAccumulatedCloud();
     }
   }
 
@@ -202,15 +202,16 @@ private:
   void checkAndSave()
   {
     if (!save_triggered_) {
-      saveAccumulatedCloud(true);
+      saveAccumulatedCloud();
+      RCLCPP_INFO(this->get_logger(), "Shutting down the node.");
+      rclcpp::shutdown();
     }
   }
 
   /**
    * @brief Writes the accumulated point cloud to disk in a single PCD file.
-   * \param do_shutdown if true, calls rclcpp::shutdown() after saving.
    */
-  void saveAccumulatedCloud(bool do_shutdown = true)
+  void saveAccumulatedCloud()
   {
     if (save_triggered_) {
       return;
@@ -242,10 +243,6 @@ private:
         "Saved %u points to %s",
         accumulated_cloud_.width * accumulated_cloud_.height,
         filename.c_str());
-    }
-    if (do_shutdown) {
-      RCLCPP_INFO(this->get_logger(), "Shutting down the node.");
-      rclcpp::shutdown();
     }
   }
 };
--- a/pcl_ros/tools/combined_pointcloud_to_pcd.cpp
+++ b/pcl_ros/tools/combined_pointcloud_to_pcd.cpp
@@ -41,15 +41,17 @@
  * into a fixed frame.
  */
 
-#include <pcl/common/io.h>
 #include <pcl/io/pcd_io.h>
-#include <pcl/point_types.h>
 #include <pcl_conversions/pcl_conversions.h>
 
+#include <chrono>
+#include <sstream>
+#include <string>
+
+#include <pcl_ros/transforms.hpp>
 #include <rclcpp/rclcpp.hpp>
 #include <rclcpp_components/register_node_macro.hpp>
 #include <sensor_msgs/msg/point_cloud2.hpp>
-#include <tf2_eigen/tf2_eigen.hpp>
 #include <tf2_ros/buffer.hpp>
 #include <tf2_ros/transform_listener.hpp>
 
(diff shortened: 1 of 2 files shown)


remco-r (approved):
Tested in jazzy on ubuntu 24 with synthetic and real data.
* Nicely aligns architecture with `pointcloud_to_pcd`
* Works in general, input fields are present in output cloud.
* Also tested with full high resolution ouster cloud and fixed frame map. The node correctly projected 5.000.000 points into map frame.

@Rayman / @SteveMacenski This one looks good to me.

**Speficic tests**

✅ pcl::PCLPointCloud2::concatenate replaces manual buffer append, adding schema validation (skips clouds with mismatched fields or endianness)
_Different point schemas are rejected with warning_

✅ is_dense propagated as logical AND across accumulated clouds:
_works with dense and non dense clouds. output cloud contains nans of input cloud._

☐ tf2_ros removed (transitively provided by pcl_ros_tf)
_For me to make build work in jazzy, had to add linking to `tf2_ros::tf2_ros` again, and mitigation of deprecation of `ament_dependencies`, while in jazzy pcl_conversions is header only. I did not test on rolling_

✅ Fixed filename bug: prefix "combined_" was producing combined_combined_*.pcd
_woks, and prefix is prepended e.g. abc_combined_1782375648325389781.pcd_

✅ rclcpp::shutdown() no longer called from destructor (passed via do_shutdown flag to avoid UB)
_Both save on shutdown, and `save_timer_sec` work. `save_timer_sec` shuts down node after saving._

pcl_ros/tools/combined_pointcloud_to_pcd.cpp
```diff
@@ -251,14 +202,15 @@ class CombinedPointCloudToPCD : public rclcpp::Node
...
+      saveAccumulatedCloud(true);
     }
   }
 
   /**
    * @brief Writes the accumulated point cloud to disk in a single PCD file.
+   * \param do_shutdown if true, calls rclcpp::shutdown() after saving.
    */
-  void saveAccumulatedCloud()
+  void saveAccumulatedCloud(bool do_shutdown = true)
```
Rayman: Why did you add a parameter? We have a member called `save_on_shutdown_` with the same information?

pcl_ros/tools/combined_pointcloud_to_pcd.cpp
```diff
@@ -267,57 +219,34 @@ class CombinedPointCloudToPCD : public rclcpp::Node
 
     // Create a filename
     std::stringstream ss;
-    ss << prefix_ << "combined_" << this->now().seconds() << ".pcd";
+    ss << prefix_ << "combined_" << this->now().nanoseconds() << ".pcd";
```
Rayman, line 223: Why did you make this change?

Rayman (commented):
I like the idea of this PR, thank you!

pcl_ros/tools/combined_pointcloud_to_pcd.cpp
kubojion: I dded it to distinguish the two save paths: the timer path should save and then stop the node, while the destructor path is already running during shutdown and should only write the file. `save_on_shutdown_` controls whether the destructor saves at all, but it does not tell `saveAccumulatedCloud()` whether it should initiate `rclcpp::shutdown()`.
I agree the parameter is confusing. I can change this so `saveAccumulatedCloud()` only writes the file, and `checkAndSave()` calls `rclcpp::shutdown()` after saving.

pcl_ros/tools/combined_pointcloud_to_pcd.cpp
kubojion: This is to avoid formatting a double from `seconds()` into the filename. nanoseconds gives an exact integer timestamp.

kubojion:
@Mergifyio queue

pcl_ros/tools/combined_pointcloud_to_pcd.cpp
Rayman: Thank you for the explanation. I think that makes more sense indeed.
- Change saveAccumulatedCloud to save only
- Then call rclcpp::shutdown in the parent function (checkAndSave)
- This allows you to reuse saveAccumulatedCloud in the destructor

kubojion: Done! `saveAccumulatedCloud()` now only saves the file, `checkAndSave()` calls `rclcpp::shutdown()` after saving, and the destructor reuses `saveAccumulatedCloud()`.

Rayman (approved):

Rayman:
The diff applied cleanly on jazzy (except for the CMakeLists.txt) so I'm going to try to backport: https://github.com/ros-perception/perception_pcl/pull/537

PR #532, opened 2026-05-27
Replace transform exceptions with the base tf2::TransformException

alireza-moayyedi:
Currently the tf2 exceptions are explicitly set to `tf2::LookupException` and `tf2::ExtrapolationException`. However, there could be more exceptions as mentioned in https://github.com/ros2/geometry2/blob/rolling/tf2/include/tf2/exceptions.hpp.

I believe it is better to use the base `tf2::TransformException` instead of explicitly adjusting the list every time.

### DIFF
--- a/pcl_ros/src/transforms.cpp
+++ b/pcl_ros/src/transforms.cpp
@@ -78,10 +78,7 @@ transformPointCloud(
       tf_buffer.lookupTransform(
       target_frame, in.header.frame_id, tf2_ros::fromMsg(
         in.header.stamp), tf2::Duration(std::chrono::seconds(1)));
-  } catch (tf2::LookupException & e) {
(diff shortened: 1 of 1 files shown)


Rayman:
I'm for this change 👍 

@SteveMacenski do we backport this to jazzy? It could be a bug, but it slightly changes behavior. I think we should backport it

Rayman (approved):

alireza-moayyedi:
Would be great if we can backport it to Jazzy!

SteveMacenski (approved):

SteveMacenski:
I think we should backport it too. Please open a PR :-)

alireza-moayyedi:
Backport PR made:
https://github.com/ros-perception/perception_pcl/pull/533

Issue #534, opened 2026-06-01
Cross-compile error: export libraries with absolute paths


kaildy:
### When cross-compile，error occured like this：

<img width="1080" height="591" alt="Image" src="https://github.com/user-attachments/assets/4ba7dd9c-3594-4369-b4b6-cdcefdab8f45" />

### file is： /opt/sysroot/opt/ros/jazzy/share/pcl_ros/cmake/export_pcl_rosExport.cmake，like this：

<img width="1080" height="262" alt="Image" src="https://github.com/user-attachments/assets/37929ad0-7b02-4f46-a7a5-85d06ef0a6fb" />

<img width="1080" height="420" alt="Image" src="https://github.com/user-attachments/assets/2d50fbc8-5c7e-4883-861f-4aa4f4096e2c" />

/opt/sysroot is the dir cross-compile toolchain dir，using this script can fix：

`cd /opt/sysroot/opt/ros/jazzy/share/pcl_ros/cmake
sudo cp export_pcl_rosExport.cmake export_pcl_rosExport.cmake.bk
sudo sed -i "s|/usr/include|/opt/sysroot/usr/include|g" export_pcl_rosExport.cmake
sudo sed -i "s|/usr/lib|/opt/sysroot/usr/lib|g" export_pcl_rosExport.cmake
sudo sed -i "s|/opt/ros/jazzy|/opt/sysroot/opt/ros/jazzy|g" export_pcl_rosExport.cmake

cd /opt/sysroot/opt/ros/jazzy/share/nav2_costmap_2d/cmake
sudo cp export_nav2_costmap_2dExport.cmake export_nav2_costmap_2dExport.cmake.bk
sudo sed -i "s|/usr/include|/opt/sysroot/usr/include|g" export_nav2_costmap_2dExport.cmake
sudo sed -i "s|/usr/lib|/opt/sysroot/usr/lib|g" export_nav2_costmap_2dExport.cmake
sudo sed -i "s|/opt/ros/jazzy|/opt/sysroot/opt/ros/jazzy|g" export_nav2_costmap_2dExport.cmake`

Rayman:
What is exactly the bug in this repository? Can you send a PR so we can discuss in more detail?

PR #535, opened 2026-06-02
fix cross compile

kaildy:
fix cross compile error, for detail see: https://github.com/ros-perception/perception_pcl/issues/534

### DIFF
--- a/pcl_ros/CMakeLists.txt
+++ b/pcl_ros/CMakeLists.txt
@@ -23,27 +23,29 @@ find_package(tf2 REQUIRED)
 find_package(tf2_geometry_msgs REQUIRED)
 find_package(tf2_ros REQUIRED)
 
-set(dependencies
-  pcl_conversions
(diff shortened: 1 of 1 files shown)


Rayman:
The way we normally work is that you first open a PR to rolling, and then we check if it can be backported to jazzy and humble.

In rolling we removed `ament_target_dependencies` already. Does that resolve your issue?

kaildy:
> The way we normally work is that you first open a PR to rolling, and then we check if it can be backported to jazzy and humble.
> 
> In rolling we removed `ament_target_dependencies` already. Does that resolve your issue?

We have not tested the rolling branch. The current environment is installed through the apt command (sudo apt install ros-jazzy-pcl-ros). We manually modified the export_pcl_rosExport.cmake file, and cross-compilation can now run normally.

Removing ament_target_dependencies should fix our issue. We'll test it later.

PR #540, opened 2026-07-09
Use the new TransformListener constructor

Rayman:
This fixes a deprecation warning. See https://github.com/ros2/geometry2/pull/714. This constructor is available in tf2_ros 0.45.0  or higher,  so lyrical or newer.


Rayman:
I don't want to create a separate branch for kilted. So let's see if the new constructor can be backported.

EDIT: It can't, so let's not merge this for now

